PR #16, opened 2018-06-17
upgraded to Bitmovin 2.10.2 and Exoplayer 2.8.1

zachmanc:
- Upgraded to Bitmovin 2.10.2 and Exoplayer 2.8.1
  - Two event listeners I was using are now deprecated. Switched to using `AnalyticsListener`
- Fixed bug in Exoplayer adapter where positions were in seconds instead of milliseconds. When fixing this bug, it caused initial positions to be small negative numbers (-25ms for example). Put in a check to set position to zero if it is negative

### DIFF
--- a/collector/src/main/java/com/bitmovin/analytics/adapters/ExoPlayerAdapter.java
+++ b/collector/src/main/java/com/bitmovin/analytics/adapters/ExoPlayerAdapter.java
@@ -1,5 +1,7 @@
 package com.bitmovin.analytics.adapters;
 
+import android.net.NetworkInfo;
+import android.support.annotation.Nullable;
 import android.util.Log;
 import android.view.Surface;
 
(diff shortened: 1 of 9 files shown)


collector/src/main/java/com/bitmovin/analytics/adapters/ExoPlayerAdapter.java
```diff
@@ -82,7 +83,11 @@ private long getPosition() {
             Timeline.Period firstPeriodInWindow = new Timeline.Period();
             if (firstPeriodInWindowIndex >= 0 && firstPeriodInWindowIndex < timeline.getPeriodCount()) {
                 timeline.getPeriod(firstPeriodInWindowIndex, firstPeriodInWindow);
-                return (exoplayer.getCurrentPosition() - firstPeriodInWindow.getPositionInWindowMs()) / 1000;
+                long position = (exoplayer.getCurrentPosition() - firstPeriodInWindow.getPositionInWindowMs());
+                if (position < 0) {
```
matamegger: Any clue why the calculated position can be negative?

zachmanc: I am not sure. I debugged some more this morning and can not find a pattern. Most of the time it doesnt happen, but when it does it is very small values that use to get rounded to 0 when divided by 0. Potentially an exoplayer issue?

PR #30, opened 2019-03-04
Feature/codec information

rolandgriesser:
- Added `audioCodec` and `videoCodec` in outgoing payload (https://github.com/bitmovin/analytics-issues/issues/398)
- Changed `title` property in payload to `videoTitle` (https://github.com/bitmovin/analytics-issues/issues/430)
- Fixed a problem where the the collector reported the player version it was compiled against in some cases, not the version in use (both Exoplayer and Bitmovin Player) (https://github.com/bitmovin/analytics-issues/issues/428)

### DIFF
--- a/collector-bitmovin-player/build.gradle
+++ b/collector-bitmovin-player/build.gradle
@@ -35,8 +35,10 @@ dependencies {
     androidTestImplementation 'com.android.support.test:runner:1.0.1'
     androidTestImplementation 'com.android.support.test.espresso:espresso-core:3.0.1'
 
-    api 'com.bitmovin.analytics:collector:1.4.2'
-    api 'com.bitmovin.player:playercore:2.+'
+    //api 'com.bitmovin.analytics:collector:1.4.2'
+    api project(':collector')
+    //api 'com.bitmovin.player:playercore:2.+'
+    api 'com.bitmovin.player:playercore:2.19.1-SNAPSHOT'
 
 }
 
--- a/collector-bitmovin-player/build.gradle
+++ b/collector-bitmovin-player/build.gradle
@@ -3,7 +3,7 @@ apply plugin: 'com.jfrog.artifactory'
 apply plugin: 'maven-publish'
 
 def packageName = 'com.bitmovin.analytics'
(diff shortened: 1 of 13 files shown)


collector-bitmovin-player/build.gradle
```diff
@@ -35,8 +35,8 @@ dependencies {
     androidTestImplementation 'com.android.support.test:runner:1.0.1'
     androidTestImplementation 'com.android.support.test.espresso:espresso-core:3.0.1'
 
-    api 'com.bitmovin.analytics:collector:1.4.2'
-    api 'com.bitmovin.player:playercore:2.+'
+    api 'com.bitmovin.analytics:collector:1.4.3'
+    implementation 'com.bitmovin.player:playercore:2.20.1'
```
Tigraine: I don't think this is good - as it will break anyone running 2.21 right?

@matamegger @hawk23 any idea how we can allow a user to select a newer version of bitmovin player (without being locked to this specific version) without updating his version to latest?
The issue with `2.+` was that it would automatically select the latest version all the time

collector-exoplayer/build.gradle
```diff
@@ -35,8 +35,8 @@ dependencies {
     androidTestImplementation 'com.android.support.test:runner:1.0.1'
     androidTestImplementation 'com.android.support.test.espresso:espresso-core:3.0.1'
 
-    api 'com.bitmovin.analytics:collector:1.4.2'
-    api 'com.google.android.exoplayer:exoplayer:2.8.+'
+    api 'com.bitmovin.analytics:collector:1.4.3'
+    implementation 'com.google.android.exoplayer:exoplayer:2.8.0'
```
Tigraine: Same here - I think this will cause all kinds of breakage

collector-bitmovin-player/build.gradle
hawk23: Let's discuss this offline, I think I don't fully understand the question.

hawk23: As discussed offline with @rolandgriesser that should not be a problem.

PR #38, opened 2019-05-03
Feature/videotime start end on heartbeat

Tigraine:
Fixes https://github.com/bitmovin/analytics-issues/issues/591


collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
```diff
@@ -214,6 +212,13 @@ public void sendEventData(EventData data) {
     this.eventDataDispatcher.add(data);
   }
 
+  public long getPosition() {
+    if(playerAdapter == null) {
+      return 0;
```
Tigraine: Is this safe to always assume position 0? or is this only used during startup?

rolandgriesser: The `playerAdapter` will be set on `attachPlayer`. So as long as there's no player attached, oit should be safe to assume that

PR #45, opened 2019-06-12
Hotfix/licensing npe

Tigraine:
### Work
- Fixes: https://github.com/bitmovin/analytics-issues/issues/660
- Description: A NullPointerException could occur if the licensing backend did respond with a malformed message that doesn't contain a json response body with a `message` payload. (The logmessage was calling `getMessage()` without checking nullability.

### Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/collector/src/main/java/com/bitmovin/analytics/utils/LicenseCall.java
+++ b/collector/src/main/java/com/bitmovin/analytics/utils/LicenseCall.java
@@ -37,19 +37,20 @@ public class LicenseCall {
 
             @Override
             public void onResponse(Call call, Response response) throws IOException {
-                if (response != null) {
-                    LicenseResponse licenseResponse = DataSerializer.deserialize(response.body().string(), LicenseResponse.class);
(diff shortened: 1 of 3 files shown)


collector/src/main/java/com/bitmovin/analytics/utils/LicenseCall.java
```diff
@@ -31,25 +31,32 @@ public void authenticate(final LicenseCallback callback) {
...
+
+                LicenseResponse licenseResponse = DataSerializer.deserialize(response.body().string(), LicenseResponse.class);
+                if (licenseResponse == null) {
                     Log.d(TAG, "License call was denied without providing a response body");
+                    callback.authenticationCompleted(false);
+                    return;
                 }
-                callback.authenticationCompleted(false);
+
+                if (!licenseResponse.getStatus().equals("granted")) {
```
zachmanc: could getStatus() be null?

PR #51, opened 2019-08-13
Feature/player bundling

mschoenlaub:
### Work
- Fixes: Prepares Android collector for inclusion in Player SDK
- Description:
This PR removes Application context from the config, so that Player SDK can actually use the config. Code around licensing is modifed to allow update of the used license key.

### Checklist

- [X] Updated CHANGELOG.md


mschoenlaub:
force-push was due to merge conflicts with in-between merged PRs

collector/src/main/java/com/bitmovin/analytics/data/SimpleEventDataDispatcher.java
```diff
@@ -1,53 +1,61 @@
...
-        this.httpClient = new HttpClient(config.getContext(), BitmovinAnalyticsConfig.analyticsUrl);
+        this.httpClient = new HttpClient(context, config.getAnalyticsUrl());
         this.config = config;
         this.callback = callback;
+        this.context = context;
     }
 
     @Override
-    synchronized public void authenticationCompleted(boolean success) {
+    synchronized public void authenticationCompleted(boolean success, String key) {
```
Tigraine: This method should probably also be renamed.

collector/src/main/java/com/bitmovin/analytics/license/DefaultLicenser.java
```diff
@@ -0,0 +1,17 @@
+package com.bitmovin.analytics.license;
+
+public class DefaultLicenser implements OnAuthCompleted, Licenser {
```
Tigraine: `Licenser` is more of a `go` name than a Java name. A better name for Java would be a `LicenseProvider` or smth similar

collector/src/main/java/com/bitmovin/analytics/utils/HttpClient.java
```diff
@@ -26,7 +26,11 @@ public HttpClient(Context context, String url) {
         this.context = context;
     }
 
+
     public void post(String postBody, final Callback callback) {
+        if (client == null) {
```
Tigraine: Why do we do this here? Is there any benefit to late-initializing the `OkHTTPClient`

collector/src/main/java/com/bitmovin/analytics/data/SimpleEventDataDispatcher.java
```diff
@@ -61,7 +69,7 @@ public void disable() {
     @Override
     public void add(EventData eventData) {
         eventData.setSequenceNumber(this.sampleSequenceNumber++);
-
+        eventData.setKey(licenser.getAnalyticsLicense());
```
Tigraine: I don't see where the Dispatcher is disabled if no license was returned by the Licensing? If I'm reading this correctly it would send out a bunch of samples with `null` as `key` and our backend would obviously reject these requests.

collector/src/main/java/com/bitmovin/analytics/data/SimpleEventDataDispatcher.java
```diff
@@ -1,53 +1,61 @@
...
+        this.httpClient = new HttpClient(context, config.getAnalyticsUrl());
         this.config = config;
         this.callback = callback;
+        this.context = context;
     }
 
     @Override
-    synchronized public void authenticationCompleted(boolean success) {
+    synchronized public void authenticationCompleted(boolean success, String key) {
         if (success) {
```
Tigraine: This was obviously missing before but now it will happen more often: We need to clear the queue and not add new things to the queue. Otherwise this is a memory leak

collector/src/main/java/com/bitmovin/analytics/utils/HttpClient.java
mschoenlaub: Because of how Player unit Tests work. :)

collector/src/main/java/com/bitmovin/analytics/license/DefaultLicenser.java
mschoenlaub: Agreed.

collector/src/main/java/com/bitmovin/analytics/data/SimpleEventDataDispatcher.java
Tigraine: This is already handled by the `detach` method.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- `BitmovinAnalyticsConfig` is serializable now. `Context` is now passed seperately where needed.
```
Tigraine: I don't think this should go into the fixed category as it wasn't really a bug right?

You can also add a `# Internal` category for changes that don't go into the public changelog on the website

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
```diff
@@ -1,228 +1,235 @@
...
+
+    /**
+     * Bitmovin Analytics
+     *
+     * @param bitmovinAnalyticsConfig {@link BitmovinAnalyticsConfig}
+     */
+    public BitmovinAnalytics(BitmovinAnalyticsConfig bitmovinAnalyticsConfig, Context context) {
+        this.context = context;
+        this.bitmovinAnalyticsConfig = bitmovinAnalyticsConfig;
+        this.licenser = new DefaultLicenseProvider();
```
Tigraine: it's still called `licenser` here :)

collector/src/main/java/com/bitmovin/analytics/utils/HttpClient.java
```diff
@@ -17,7 +17,7 @@
     public static final MediaType JSON
             = MediaType.parse("application/json; charset=utf-8");
     private static final String TAG = "HttpClient";
-    private final OkHttpClient client = new OkHttpClient();
+    private OkHttpClient client = null;
```
Tigraine: we should make a note somewhere to discuss this with the player team and urge them to fix this. It's ok here for now but this smells like something that will bite us in the ass :D in the end.

Does it make sense to move the HTTP Client into a late-init singleton implementation in a separate class?

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalyticsConfig.java
```diff
@@ -1,44 +1,118 @@
...
+    private String key;
+    private String title;
     private String path;
+    private String playerKey;
+    private PlayerType playerType;
+    private String videoId;
     private Context context;
-    private int heartbeatInterval = 59700;
 
+    public static final Creator<BitmovinAnalyticsConfig> CREATOR = new Creator<BitmovinAnalyticsConfig>() {
```
Tigraine: What does this do? 
I don't see it being used anywhere

collector/src/main/java/com/bitmovin/analytics/license/LicenseCall.java
```diff
@@ -15,54 +19,66 @@
...
+    private String getLicenseKeyForChecking(){
+        String key = config.getKey();
+        if (key == null || key.isEmpty()){
+            key = licenseProvider.getAnalyticsLicense();
+        }
+        return key;
+    }
+
+    public void authenticate(final OnLicenseValidated callback) {
+        final LicenseCallData data = new LicenseCallData();
```
Tigraine: Not sure if that makes sense to have the `key` from the config here on the `LicenseCallData`. Seems like "re-using" a concept but introduces more complexity IMO. Maybe we can try to refactor this together in a short pairing session to see if we can make it more elegant?

analyticsexample/src/main/java/com/bitmovin/exoplayeranalyticsexample/MainActivity.java
```diff
@@ -93,8 +93,7 @@ private void createPlayer() {
             bitmovinAnalyticsConfig.setHeartbeatInterval(59700);
 
             //Step 3: Create Analytics Colelctor
```
peterpf: There's a typo in the comment

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
```diff
@@ -1,228 +1,246 @@
...
+    }
+
+    /**
+     * Bitmovin Analytics
+     *
+     * @param bitmovinAnalyticsConfig {@link BitmovinAnalyticsConfig}
+     */
+    @Deprecated
+    public BitmovinAnalytics(BitmovinAnalyticsConfig bitmovinAnalyticsConfig) {
+        this(bitmovinAnalyticsConfig, bitmovinAnalyticsConfig.getContext());
```
peterpf: Check whether `getContext` is null

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalyticsConfig.java
mschoenlaub: It is used by Android  (e.g. see https://medium.com/@nikhildhyani365/understand-parcelable-in-android-27ce420d695b, https://www.vogella.com/tutorials/AndroidParcelable/article.html)

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinPlayerCollector.java
```diff
@@ -12,13 +14,19 @@
...
-      BitmovinAnalyticsConfig bitmovinAnalyticsConfig) {
-    super(bitmovinAnalyticsConfig);
+          BitmovinAnalyticsConfig bitmovinAnalyticsConfig, Context context) {
+    super(bitmovinAnalyticsConfig, context);
+  }
+
+  @Deprecated
+  public BitmovinPlayerCollector(
+          BitmovinAnalyticsConfig bitmovinAnalyticsConfig) {
+    this(bitmovinAnalyticsConfig, bitmovinAnalyticsConfig.getContext());
```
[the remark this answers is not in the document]
mschoenlaub: Done in base class now. Should catch all cases where context could be null

collector/src/main/java/com/bitmovin/analytics/license/LicenseCall.java
mschoenlaub: Hm... yeah... We could also put this into the LicenseProvider

PR #54, opened 2019-08-29
Empty bitmovin.properties file

zachmanc:
### Work
- Fixes: Gradle fails to sync when you check out develop
- Description: Gradle references a bitmovin.properties file that is not included in the project. I added an "empty" one that allows builds to succeed 

### Checklist

- [ ] Updated CHANGELOG.md


Tigraine:
Hey cory - thanks for pointing this out.
Since I had to fix the build process for publishing anyways I fixed this with a conditional check instead of the empty file (as that would probably be problematic with our local `.gitignore` rules:

`master` and `develop` have the fix here:
https://github.com/bitmovin/bitmovin-analytics-collector-android/commit/e26772e6f7b1f070d43b9b66385c1a4461144f72

PR #56, opened 2019-09-18
AN-828 Add audio and subtitle tracking

mschoenlaub:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-725, https://bitmovin.atlassian.net/browse/AN-828
- Description: Watch for events of changing subtitles or audio track and track them in the event data. Also send a boolean, indicating if any subtitle is enabled. 

### Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
+++ b/collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
@@ -218,21 +218,17 @@ public class BitmovinSdkAdapter implements PlayerAdapter {
             data.setAudioCodec(audioQuality.getCodec());
         }
 
-        LanguageInformation languageInformation = new LanguageInformation();
-
         //Subtitle info
         SubtitleTrack subtitle = bitmovinPlayer.getSubtitle();
         if (subtitle != null && subtitle.getId() != null) {
-            languageInformation.setSubtitleLanguage(subtitle.getLanguage());
+            data.setSubtitleLanguage(subtitle.getLanguage());
         }
 
         //Audio language
         AudioTrack audioTrack = bitmovinPlayer.getAudio();
         if (audioTrack != null && audioTrack.getId() != null) {
-            languageInformation.setAudioLanguage(audioTrack.getLanguage());
+            data.setAudioLanguage(audioTrack.getLanguage());
         }
-
-        data.setLanguageInformation(languageInformation);
 
         return data;
     }
--- a/collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
+++ b/collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
@@ -16,9 +16,39 @@ import com.bitmovin.analytics.enums.PlayerType;
 import com.bitmovin.analytics.stateMachines.PlayerState;
 import com.bitmovin.analytics.stateMachines.PlayerStateMachine;
 import com.bitmovin.analytics.utils.Util;
-import com.bitmovin.player.*;
-import com.bitmovin.player.api.event.data.*;
-import com.bitmovin.player.api.event.listener.*;
+import com.bitmovin.player.BitmovinPlayer;
+import com.bitmovin.player.api.event.data.AudioChangedEvent;
+import com.bitmovin.player.api.event.data.AudioPlaybackQualityChangedEvent;
+import com.bitmovin.player.api.event.data.DroppedVideoFramesEvent;
+import com.bitmovin.player.api.event.data.ErrorEvent;
+import com.bitmovin.player.api.event.data.PausedEvent;
+import com.bitmovin.player.api.event.data.PlayEvent;
+import com.bitmovin.player.api.event.data.PlaybackFinishedEvent;
+import com.bitmovin.player.api.event.data.ReadyEvent;
+import com.bitmovin.player.api.event.data.SeekEvent;
+import com.bitmovin.player.api.event.data.SeekedEvent;
+import com.bitmovin.player.api.event.data.SourceLoadedEvent;
+import com.bitmovin.player.api.event.data.SourceUnloadedEvent;
+import com.bitmovin.player.api.event.data.StallEndedEvent;
+import com.bitmovin.player.api.event.data.StallStartedEvent;
+import com.bitmovin.player.api.event.data.SubtitleChangedEvent;
+import com.bitmovin.player.api.event.data.VideoPlaybackQualityChangedEvent;
+import com.bitmovin.player.api.event.listener.OnAudioChangedListener;
+import com.bitmovin.player.api.event.listener.OnAudioPlaybackQualityChangedListener;
+import com.bitmovin.player.api.event.listener.OnDroppedVideoFramesListener;
+import com.bitmovin.player.api.event.listener.OnErrorListener;
+import com.bitmovin.player.api.event.listener.OnPausedListener;
+import com.bitmovin.player.api.event.listener.OnPlayListener;
+import com.bitmovin.player.api.event.listener.OnPlaybackFinishedListener;
+import com.bitmovin.player.api.event.listener.OnReadyListener;
+import com.bitmovin.player.api.event.listener.OnSeekListener;
+import com.bitmovin.player.api.event.listener.OnSeekedListener;
+import com.bitmovin.player.api.event.listener.OnSourceLoadedListener;
+import com.bitmovin.player.api.event.listener.OnSourceUnloadedListener;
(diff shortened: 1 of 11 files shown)


collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
```diff
@@ -295,6 +291,30 @@ public void onStallEnded(StallEndedEvent stallEndedEvent) {
...
+                stateMachine.transitionState(PlayerState.AUDIOTRACKCHANGE, getPosition());
+                stateMachine.transitionState(originalState, getPosition());
+            }
+        }
+    };
+
+    private OnSubtitleChangedListener onSubtitleChangedListener = new OnSubtitleChangedListener() {
+        @Override
+        public void onSubtitleChanged(SubtitleChangedEvent event) {
+            Log.d(TAG, "On SubtitleChanged: " + bitmovinPlayer.getAudio().getId());
```
peterpf:
```suggestion
            Log.d(TAG, "On SubtitleChanged: " + bitmovinPlayer.getSubtitle().getId());
```

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
```diff
@@ -229,6 +229,28 @@ public void onVideoChange() {
         Log.d(TAG, "onVideoChange");
     }
 
+  @Override
+  public void onSubtitleChange() {
+    Log.d(TAG, String.format("onAudioTrackChange %s", playerStateMachine.getImpressionId()));
```
peterpf:
```suggestion
    Log.d(TAG, String.format("onSubtitleChange %s", playerStateMachine.getImpressionId()));
```

collector/src/main/java/com/bitmovin/analytics/data/HttpBackend.java
```diff
@@ -20,13 +20,13 @@ public HttpBackend(CollectorConfig config, Context context) {
 
     @Override
     public void send(EventData eventData) {
-        Log.d(TAG, String.format("Sending sample: %s (state: %s, videoId: %s, startupTime: %d, videoStartupTime: %d, buffered: %d)",
+        Log.d(TAG, String.format("Sending sample: %s (state: %s, videoId: %s, startupTime: %d, videoStartupTime: %d, buffered: %d, language: %s)",
```
peterpf:
```suggestion
        Log.d(TAG, String.format("Sending sample: %s (state: %s, videoId: %s, startupTime: %d, videoStartupTime: %d, buffered: %d, audioLanguage: %s)",
```

collector/src/main/java/com/bitmovin/analytics/data/EventData.java
```diff
@@ -70,6 +70,9 @@
     private String audioCodec;
     private List<String> supportedVideoCodecs;
     private DeviceInformation deviceInformation = new DeviceInformation();
+    private boolean subtitleEnabled = false;
```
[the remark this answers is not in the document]
mschoenlaub: Actually Gson (in contrast to Jackson) works via field reflection and seems to ignore setters/getters altogether.

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
```diff
@@ -229,6 +229,28 @@ public void onVideoChange() {
...
+    EventData data = playerAdapter.createEventData();
+    data.setState(playerStateMachine.getCurrentState().toString().toLowerCase());
+    data.setDuration(0);
```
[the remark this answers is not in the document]
rolandgriesser: AFAIK this will cause the state machine to call the `onExitState` from the `PLAYING` state - so it will send a playing-sample with the duration up to now (`onExitState` passes the duration as `timeStamp - machine.getOnEnterStateTimeStamp()`). Then it will call the `onEnterState` from the `SUBTITLECHANGE`, which will update the `enterStateTimeStamp` for succeeding state changes.
Next thing, going back to the `PLAYING` state again will cause the `onExitState` from `SUBTITLECHANGE` to be called, which will send a 0-duration sample indicating that a subtitle change occurred, and immediately enter the old state `PLAYING` again - `enterStateTimeStamp` will be updated again and the next duration will be fine again. So all good.
The only problem we're having (same as in Web): Sending a sample will take the current language from the `BitmovinAdapter` - which in this case will already be the new one. So the old `playing` sample, which had the old language, will be send with the new one. This will have to be addressed for everything else as well.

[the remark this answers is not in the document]
Tigraine: So the state transitions are actually correct and sending the "old" sample with the new language is actually a known issue :/

Tigraine:
> I tested this and it seems we have a small issue with the new audio language being sent on the "last" sample.
> 
> So basically
> 
> ```
> t 0                    1                            2
>   +----- en ---+
>                        +------- de ---- +
> ```
> 
> So imagine at t0 we start playing and at t1 we switch the audio language.
> The expected behavior is to send out a sample with `playing: 1, audioLanguage: 'en` and start a new span with audiolanguage `de` after t1 and at t2 we send another one.
> 
> The sample sequence should look like:
> 
> ```
> sample1: start: 0, end: 1, audioLanguage: en
> sample1: start 1, end 2, audioLanguage: de
> ```
> 
> What actually happens:
> 
> ```
> sample1: start: 0, end: 1, audioLanguage: de
> sample1: start 1, end 2, audioLanguage: de
> ```
> 
> It's not that extreme because of a bunch of rebuffering and other stuff that happens but the event sequence is a bit wrong.
> 
> I have added a commit to the PR with a test-stream with multiple languages that can be used to test.
> 
> I also tested this a bit with paused state and it seems we need to "suppress" a few events during the transition as when I change I often times get `qualitychange` events as well as `subtitlechange` events which don't really make any sense..
> 
> might make sense to jump on a screenshare tomorrow morning and discuss this in detail.

As discussed in the comments and after @rolandgriesser debugging through this I think this is fine. The change with the `languageInformation` change should still go in but aside from that this is great 👍 so disregard my last concern about the language sample.

collector/src/main/java/com/bitmovin/analytics/data/EventData.java
```diff
@@ -594,4 +596,14 @@ public String getAudioCodec() {
     public DeviceInformation getDeviceInformation() {
         return deviceInformation;
     }
+
+    public LanguageInformation getLanguageInformation() {
```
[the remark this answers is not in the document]
mschoenlaub: I think we did start nesting in some cases (e.g. DeviceInfo and some bandwidth stuff), but now the web collector - as the reference standard - does not do this for the fields in question. Obviously there is no way to change that anymore, but I would rather prefer readable DTOs and mess just a little bit with serialization than continuing adding more and more fields, getters and setters to a 609 LOC big DTO.

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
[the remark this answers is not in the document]
mschoenlaub: We should however ix this issue (maybe thoroughly thinking through state transitions once again :)

collector/src/main/java/com/bitmovin/analytics/data/EventData.kt
```diff
@@ -0,0 +1,76 @@
...
+    var isMuted = false
+    var sequenceNumber: Int = 0
+    val platform: String = "android"
+    var videoCodec: String? = null
+    var audioCodec: String? = null
+    var supportedVideoCodecs: List<String>? = null
+    val deviceInformation = DeviceInformation()
+    var subtitleEnabled: Boolean = false
+    var subtitleLanguage: String? = null
+    set(value) {
```
rolandgriesser: Still would prefer to not compute that.
Reasoning: Maybe there's a case where we can't get the language but subtitles are still enabled? I know, shouldn't be the case, but let's the Adapter make that decision.
Otherwise, I'd switch to a getter on the `subtitleEnabled`.
Reasoning:
Imagine deserialization working this way: `subtitleLanguage` is deserialized and set, `subtitleEnabled` will be set to true, and then someone passes `subtitleEnabled` as well - value will be overridden. Having no setter on `subtitleEnabled` would prevent this. Also, I'm not sure if on an initial assignment through the constructor, the setter will be called. Usually you invert the dependency here.

will let @Tigraine decide

mschoenlaub: @rolandgriesser getter on `subtitleEnabled` was my first idea too. (Mainly becaue it prohibits people doing stupid things like overriding a value like this :) 
@Tigraine had objections though.

Tigraine: I'd leave it at this right now as it does what we need and can be refactored for the other case (with maybe other collectors) at a laster stage

PR #58, opened 2019-10-16
AN-835 Feature/exoplayer error detail

mschoenlaub:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-835
- Description: Refactored the error mapping functionality. Fix leakage of information into error message (moved this to error data). Also, a part of the stack trace is now logged, so that we can see which are the "UNKNOWN" esceptions.

### Checklist

- [x] Updated CHANGELOG.md


collector/src/main/java/com/bitmovin/analytics/data/ErrorCode.kt
```diff
@@ -18,33 +18,15 @@
...
-    }
-
     @Override
-    public String toString() {
-        return errorCode + ": " + description;
+    override fun toString() : String {
+        return errorCode.toString() + ": " + description;
     }
+
+    data class ErrorData(val msg: String, val details: Array<String> = emptyArray())
```
Tigraine: Can this be pushed one level higher? Reads a bit "choppy" `ErrorCode.ErrorData`?

collector-exoplayer/src/main/java/com/bitmovin/analytics/exoplayer/ExceptionMapper.kt
```diff
@@ -0,0 +1,58 @@
...
+                errorCode.errorData = ErrorCode.ErrorData(error.message
+                        ?: "", error.topOfStacktrace)
+            }
+        }
+        return errorCode
+    }
+
+    override fun map(throwable: Throwable): ErrorCode {
+        return when (throwable) {
+            is ExoPlaybackException -> map(throwable)
```
peterpf: how do i know which `map` will be called? Please make it more readable

collector-exoplayer/src/test/java/com/bitmovin/analytics/error/ErrorCodeTest.kt
```diff
@@ -0,0 +1,23 @@
+package com.bitmovin.analytics.error
```
peterpf: Those tests don't check the errorCode :)

collector/src/test/java/com/bitmovin/analytics/error/ErrorDataTest.kt
```diff
@@ -0,0 +1,23 @@
+package com.bitmovin.analytics.error
+
+import com.bitmovin.analytics.data.ErrorData
+import com.bitmovin.analytics.utils.DataSerializer
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+
+class UtilTest {
```
peterpf:
```suggestion
class ErrorDataTest {
```
This is the same file as `collector-exoplayer/src/test/java/com/bitmovin/analytics/error/ErrorDataTest.kt` except that the className is different.
Is it necessary to duplicate those?

mschoenlaub: Nope, was a leftover. thx for pointing out

PR #59, opened 2019-10-24
AN-939 added Device Orientation for Android collector

sabe25:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-939
- Description: added device orientation for android collector

### Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Development
 
 ### Added
-
 - sending information event for `device orientation`
 
 ## v1.9.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Development
 
+- sending information event for `device orientation`
+
 ## v1.9.0
 ### Fixed
 - Error message contained too much data, so that grouping is not useful
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Development
 
+### Added
+
 - sending information event for `device orientation`
 
 ## v1.9.0
--- a/collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
+++ b/collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
@@ -1,8 +1,13 @@
 package com.bitmovin.analytics;
 
+import static android.content.Context.SENSOR_SERVICE;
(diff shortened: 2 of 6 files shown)


collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
```diff
@@ -86,6 +92,31 @@ public void detachPlayer() {
         eventDataDispatcher.disable();
     }
 
+    private void addDeviceOrientationListener() {
+
+        SensorEventListener m_sensorEventListener = new SensorEventListener() {
```
Tigraine: We don't use `_` in the variable names. We try to keep to the [google code style ](https://google.github.io/styleguide/javaguide.html)

A better name would just be `sensorEventListener`

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
```diff
@@ -86,6 +92,31 @@ public void detachPlayer() {
...
+                    isPortrait = !isPortrait;
+                }
+            }
+            @Override
+            public void onAccuracyChanged(Sensor sensor, int accuracy) {
+                // No state change here
+            }
+        };
+
+        SensorManager sm = (SensorManager) context.getSystemService(SENSOR_SERVICE);
```
Tigraine: Is this cast guaranteed to always succeed? What if `getSystemService` returns null?

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
Tigraine: Also a good trick in IntelliJ to save on typing is to just select the expression that generates a variable and then use `Refactor -> Extract -> Variable` as IntelliJ will then infer the type and also suggest a usually semi-useful name. 

image

The shortut on my machine is `Option+Command+V`

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Development
 
+### Added
+
```
Tigraine:
```suggestion
```

collector/src/main/java/com/bitmovin/analytics/utils/Util.java
```diff
@@ -87,4 +87,18 @@ public static boolean isMimeTypeSupported(String mimeType) {
         }
         return false;
     }
+
+    public static String getDeviceOrientation(Context context) {
+//        Display display = ((WindowManager) context.getSystemService(Context.WINDOW_SERVICE)).getDefaultDisplay();
```
Tigraine: Do we need this comment here?

Tigraine:
Small thing on the comment otherwise LGTM

PR #60, opened 2019-10-29
Implement Analytics AdAdapter for Bitmovin Player on Android

sabe25:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-909
- Description: Finished implementation of BitmovinSdkAdAdapter
added test setup for Ads to Bitmovin Example App

### Checklist

- [x] Updated CHANGELOG.md


collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
```diff
@@ -0,0 +1,101 @@
...
+            adAnalytics.onAdStarted(adFactory.FromPlayerAd(it.ad!!))
+        }
+    }
+
+    private val onAdFinishedListener = OnAdFinishedListener {
+        adAnalytics.onAdFinished()
+    }
+
+    private val onAdBreakStartedListener = OnAdBreakStartedListener {
+        if (it.adBreak != null){
```
Tigraine: This looks honestly a bit weird - why would we get a adBreakStarted if the adBreak is null?
Also we should do a early return here. 

```
if (it.adBreak == null) {
  return
}}
// adAnalytics.onAd..
```

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
```diff
@@ -0,0 +1,101 @@
...
+    private val onAdBreakFinishedListener = OnAdBreakFinishedListener {
+        adAnalytics.onAdBreakFinished()
+    }
+
+    private val onAdClickedListener = OnAdClickedListener {
+        adAnalytics.onAdClicked(it.clickThroughUrl)
+    }
+
+    private val onAdErrorListener = OnAdErrorListener {
+        if(it.adConfiguration != null && it.adConfiguration!! is com.bitmovin.player.model.advertising.ima.ImaAdBreak)
```
Tigraine: Same here - early return and better type cast clarifies the code

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
Tigraine: This code fixes the unsafe `!!` null conversion
```
val imaAdBreak = it.adBreak as? com.bitmovin.player.model.advertising.ima.ImaAdBreak
                ?: return@OnAdBreakStartedListener
adAnalytics.onAdBreakStarted(adBreakFactory.FromPlayerImaAdBreak(imaAdBreak))
```

This way we exploit kotlin's null guarantees

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
```diff
@@ -0,0 +1,101 @@
...
+        {
+            adAnalytics.onAdError(
+                    adBreakFactory.FromPlayerImaAdBreak(it.adConfiguration!! as com.bitmovin.player.model.advertising.ima.ImaAdBreak),
```
Tigraine: We should be able to remove the `!!` right?

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+
+    private val onAdSkippedListener = OnAdSkippedListener {
+        adAnalytics.onAdSkipped()
+    }
+
+    private val onAdManifestLoadedListener = OnAdManifestLoadedListener {
+        if (it.adBreak != null){
+            adAnalytics.onAdManifestLoaded(adBreakFactory.FromPlayerImaAdBreak(it.adBreak!! as com.bitmovin.player.model.advertising.ima.ImaAdBreak), it.downloadTime)
```
Tigraine: Same here early return and remove `!!`

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
```diff
@@ -0,0 +1,101 @@
...
+        bitmovinPlayer.addEventListener(onAdErrorListener)
+        bitmovinPlayer.addEventListener(onAdSkippedListener)
+        bitmovinPlayer.addEventListener(onAdManifestLoadedListener)
+        bitmovinPlayer.addEventListener(onPlayListener)
+        bitmovinPlayer.addEventListener(onPausedListener)
+        bitmovinPlayer.addEventListener(onAdQuartileListener)
+    }
+
+    override fun release() {
+
```
Tigraine: I think we need to remove all eventListeners that we added with `init` here

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/utils/AdTagFactory.kt
```diff
@@ -0,0 +1,17 @@
+package com.bitmovin.analytics.bitmovin.player.utils
+
+import com.bitmovin.analytics.ads.AdTagType
+import com.bitmovin.player.model.advertising.AdTag
+
+class AdTagFactory {
+
+    fun FromPlayerAdTag(adTag: AdTag): AdTagType? {
```
Tigraine: Do we really want to make this nullable?

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -0,0 +1,225 @@
...
+
+    fun onAdStarted(ad: Ad) {
+        if (ad.isLinear != true) {
+            return
+        }
+
+        this.resetActiveAd()
+        this.activeAdSample = AdSample(ad = ad)
+        this.currentTime = null
+        this.activeAdSample!!.adStartupTime = if (this.adStartupTimestamp != null) Util.getTimeStamp() - this.adStartupTimestamp!! else null
```
Tigraine: `!!` will throw NPE if not set. We should guard against it

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -0,0 +1,225 @@
...
+
+        this.startAd(this.activeAdSample!!)
+    }
+
+    fun onAdFinished() {
+        if (this.activeAdBreak == null || this.activeAdSample == null) {
+            return
+        }
+
+        val adSample = this.activeAdSample!!.copy()
```
Tigraine: The `!!` could be removed

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -0,0 +1,225 @@
...
+    fun onAdBreakFinished() {
+        this.resetActiveAd()
+        this.activeAdBreak = null
+    }
+
+    fun onAdClicked(clickThroughUrl: String?) {
+        if (this.activeAdSample == null) {
+            return
+        }
+        this.activeAdSample!!.ad.clickThroughUrl = clickThroughUrl
```
Tigraine: Don't think this is required as the compiler can check the nullability

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -0,0 +1,225 @@
...
+        this.activeAdSample!!.clicked = 1
+        this.activeAdSample!!.clickPosition = this.currentTime
+        this.activeAdSample!!.clickPercentage = Util.calculatePercentage(this.activeAdSample!!.clickPosition, this.activeAdSample!!.ad.duration)
+    }
+
+    fun onAdError(adBreak: AdBreak, code: Int?, message: String?) {
+        val adSample: AdSample
+        //todo change to references here
+        if (this.activeAdSample != null &&
```
Tigraine: Nullability

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -0,0 +1,225 @@
...
+            this.isPlaying = false
+        }
+    }
+
+    fun onAdSkipped() {
+        if (this.activeAdBreak == null || this.activeAdSample == null) {
+            return;
+        }
+
+        val adSample = this.activeAdSample!!.copy()
```
Tigraine: `!!`

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -0,0 +1,225 @@
...
+        this.resetActiveAd()
+        this.completeAd(this.activeAdBreak!!, adSample, adSample.skipPosition)
+    }
+
+    fun onAdQuartile(quartile: AdQuartile) {
+        if (this.activeAdSample == null) {
+            return
+        }
+        when {
+            quartile === AdQuartile.FIRST_QUARTILE -> this.activeAdSample!!.quartile1 = 1
```
Tigraine: `!!`

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -0,0 +1,225 @@
...
+        adSample.started = 1
+        adSample.timePlayed = 0
+        adSample.timeInViewport = 0
+        adSample.adPodPosition = this.adPodPosition
+        val timestamp = Util.getTimeStamp()
+        this.beginPlayingTimestamp = timestamp
+        // TODO this.enterViewportTimestamp = this.isContainerInViewport() ? timestamp : undefined;
+        this.isPlaying = true
+        this.currentTime = 0
+        this.adPodPosition++
```
Tigraine: I guess this isn't finished yet?

collector/src/main/java/com/bitmovin/analytics/ads/Ad.kt
```diff
@@ -0,0 +1,35 @@
+package com.bitmovin.analytics.ads
+
+data class Ad(var isLinear: Boolean? = null,
```
Tigraine: We should re-visit if really every single one of these can be nullable or not. Might make sense to have some non-nullables here that are always guaranteed to be set.. (Could also be a last step once we have the impl nailed down)

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
sabe25: Agree!

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
sabe25: if i remove the `!!` it tells me something like "impossible, because declared in a different module...

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
sabe25: i moved the casts into the mapper

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/utils/AdTagFactory.kt
sabe25: I added the AdTagType UNKNOWN and return it if incoming adTagType is not convertable

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
sabe25: two lines above there is a constructor call 
should be enough right?

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
sabe25: implemented early return

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
sabe25: compiler says no.... because it is an mutable Property we have to add it although there is obviously a null guard

collector/src/main/java/com/bitmovin/analytics/ads/Ad.kt
sabe25: Agree here, i can remove the nullable from those values directly convertable from the Ad interface of the player
we should keep nullable for every other fields which are cast depending or based on nullabel values 
What do you think?

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
```diff
@@ -14,25 +14,29 @@ import com.bitmovin.player.api.event.listener.*
  */
 class BitmovinSdkAdAdapter(val bitmovinPlayer: BitmovinPlayer, val adAnalytics: BitmovinAdAnalytics) : AdAdapter {
 
-    private val adFactory: AdFactory = AdFactory()
-    private val adBreakFactory: AdBreakFactory = AdBreakFactory()
+    private val adMapper: AdMapper = AdMapper()
```
Tigraine: 👍

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
```diff
@@ -14,25 +14,29 @@ import com.bitmovin.player.api.event.listener.*
...
     }
 
     private val onAdBreakStartedListener = OnAdBreakStartedListener {
-        if (it.adBreak != null){
-            adAnalytics.onAdBreakStarted(adBreakFactory.FromPlayerImaAdBreak(it.adBreak!! as com.bitmovin.player.model.advertising.ima.ImaAdBreak))
+        if (it.adBreak == null){
+            return@OnAdBreakStartedListener
         }
+
+        adAnalytics.onAdBreakStarted(adBreakMapper.FromPlayerAdBreak(it.adBreak!!))
```
Tigraine: Elvis operator would remove the unsafe `!!`

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/utils/AdMapper.kt
```diff
@@ -0,0 +1,72 @@
...
+        collectorAd.height = playerAd.height
+        collectorAd.id = playerAd.id
+        collectorAd.mediaFileUrl = playerAd.mediaFileUrl
+        collectorAd.clickThroughUrl = playerAd.clickThroughUrl
+        collectorAd.bitrate = playerAd.data?.bitrate
+        collectorAd.minBitrate = playerAd.data?.minBitrate
+        collectorAd.maxBitrate = playerAd.data?.maxBitrate
+        collectorAd.mimeType = playerAd.data?.mimeType
+
+        if (playerAd.data is VastAdData)
```
Tigraine: This should probably have some unit tests

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
sabe25: Agree, comes with the next commit

collector-bitmovin-player/src/test/java/com/bitmovin/analytics/bitmovin/player/utils/AdBreakMapperTest.kt
```diff
@@ -30,35 +31,49 @@ class AdBreakMapperTest {
...
+        assertThat(collectorAdBreak.scheduleTime).isNull()
+    }
+
+    @Test
+    fun FromPlayerAdConfigurationWithAdBreakShouldSetValuesOfAdBreak() {
         val collectorAdBreak = adBreakMapper.FromPlayerAdConfiguration(adBreak)
 
-        Assert.assertNotNull(collectorAdBreak.scheduleTime)
-        Assert.assertNull(collectorAdBreak.position)
+        assertThat(collectorAdBreak.scheduleTime).isNotNull()
```
Tigraine: Since you are setting it to a specific value at the `setup` it would be better to also check that value here. The test does what you intend - but it doesn't really check if the value is also correctly set - and if we were to change the data type to a not-nullable int (which defaults to 0) this test would not fail

collector-bitmovin-player/src/test/java/com/bitmovin/analytics/bitmovin/player/utils/AdMapperTest.kt
```diff
@@ -0,0 +1,70 @@
...
+        assertThat(collectorAd.bitrate).isNull()
+        assertThat(collectorAd.title).isNull()
+    }
+
+    @Test
+    fun FromPlayerAdWithLinearAdShouldSetLinearAdValues() {
+        `when`(linearAd.duration).thenReturn(10.0)
+        var collectorAd = adMapper.FromPlayerAd(linearAd)
+
+        assertThat(collectorAd.duration).isEqualTo(10000)
```
Tigraine:
```suggestion
        assertThat(collectorAd.duration).isEqualTo(TEN_SECONDS_IN_MS)
```

👍 great to check directly. 
Might make sense to extract a constant to clarify what this 10000 does exactly

collector-bitmovin-player/src/test/java/com/bitmovin/analytics/bitmovin/player/utils/AdMapperTest.kt
```diff
@@ -0,0 +1,70 @@
...
+        `when`(playerAd.data).thenReturn(null)
+        var collectorAd = adMapper.FromPlayerAd(playerAd)
+
+        assertThat(collectorAd.bitrate).isNull()
+        assertThat(collectorAd.title).isNull()
+    }
+
+    @Test
+    fun FromPlayerAdWithLinearAdShouldSetLinearAdValues() {
+        `when`(linearAd.duration).thenReturn(10.0)
```
Tigraine:
```suggestion
        TEN_SECONDS = 10.0
        TEN_SECONDS_IN_MS = TEN_SECONDS * 1000
        `when`(linearAd.duration).thenReturn(TEN_SECONDS)
```

Tigraine: obviously this won't compile as I am a kotlin noob but something like that might make the code more readable :)

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -0,0 +1,226 @@
...
+    fun onAdStarted(ad: Ad) {
+        if (!ad.isLinear) {
+            return
+        }
+
+        this.resetActiveAd()
+        this.activeAdSample = AdSample(ad = ad)
+        val activeAdSample = this.activeAdSample ?: return
+        this.currentTime = null
+        activeAdSample.adStartupTime = if (this.adStartupTimestamp != null) Util.getTimeStamp() - this.adStartupTimestamp!! else null
```
Tigraine: This is a bit hard to read imo.. clearer would be if it's less concise.
Also with your null check the smartCast should remove the need for `!!`
Maybe we can move this to a method or something - not sure - but it is a bit hard to  follow all in one  line

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -0,0 +1,226 @@
...
+
+    fun onAdClicked(clickThroughUrl: String?) {
+        val activeAdSample = this.activeAdSample ?: return
+
+        activeAdSample.ad.clickThroughUrl = clickThroughUrl
+        activeAdSample.clicked = 1
+
+        val timestamp = Util.getTimeStamp()
+        val timePlayed: Long = if (activeAdSample.timePlayed != null) activeAdSample.timePlayed!! else 0
+        activeAdSample.clickPosition = timePlayed + timestamp - this.beginPlayingTimestamp!!
```
Tigraine: I think we might be able to move all this current time calculation to some method or object inside the ad adapter. Feels like this is scattered all over :)

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -0,0 +1,226 @@
...
+        activeAdSample.clickPercentage = Util.calculatePercentage(activeAdSample.clickPosition, activeAdSample.ad.duration)
+    }
+
+    fun onAdError(adBreak: AdBreak, code: Int?, message: String?) {
+        val adSample = this.activeAdSample ?: AdSample()
+
+        if (this.activeAdSample != null &&
+                adBreak.ads.any { ad -> ad.id == this.activeAdSample!!.ad.id }) {
+            val timestamp = Util.getTimeStamp()
+            val timePlayed: Long = if (adSample.timePlayed != null) adSample.timePlayed!! else 0
```
Tigraine: I really think we should get rid of all `!!` usages if we can

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -0,0 +1,226 @@
...
+        val timePlayed: Long = if (activeAdSample.timePlayed != null) activeAdSample.timePlayed!! else 0
+        activeAdSample.clickPosition = timePlayed + timestamp - this.beginPlayingTimestamp!!
+        activeAdSample.clickPercentage = Util.calculatePercentage(activeAdSample.clickPosition, activeAdSample.ad.duration)
+    }
+
+    fun onAdError(adBreak: AdBreak, code: Int?, message: String?) {
+        val adSample = this.activeAdSample ?: AdSample()
+
+        if (this.activeAdSample != null &&
+                adBreak.ads.any { ad -> ad.id == this.activeAdSample!!.ad.id }) {
```
Tigraine: this can be changed to `this.activeAdSample?.ad?.id` would make sure we either get `id` or `null` which can still be compared to `ad.id`

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -0,0 +1,226 @@
...
+            val timestamp = Util.getTimeStamp()
+            this.beginPlayingTimestamp = timestamp
+            // TODO this.enterViewportTimestamp = timestamp
+            this.isPlaying = true
+        }
+    }
+
+    fun onPause() {
+        if (this.analytics.adAdapter != null && this.analytics.adAdapter.isLinearAdActive && this.activeAdSample != null) {
+            this.updatePlayingTime(this.activeAdSample!!)
```
Tigraine: Elvis operator?

collector-bitmovin-player/src/test/java/com/bitmovin/analytics/bitmovin/player/utils/AdBreakMapperTest.kt
```diff
@@ -0,0 +1,82 @@
...
+        val collectorAdBreak = adBreakMapper.FromPlayerAdConfiguration(adConfiguration)
+
+        assertThat(collectorAdBreak.replaceContentDuration).isEqualTo(TEN_SECONDS)
+    }
+
+    @Test
+    fun FromPlayerAdConfigurationWithAdConfigurationShouldNotSetValuesOfAdBreak(){
+        val collectorAdBreak = adBreakMapper.FromPlayerAdConfiguration(adConfiguration)
+
+        assertThat(collectorAdBreak.scheduleTime).isEqualTo(TEN_SECONDS)
```
Tigraine: Is this really in seconds? In the sample shouldn't this be a millisecond timestamp?

collector-bitmovin-player/src/test/java/com/bitmovin/analytics/bitmovin/player/utils/AdBreakMapperTest.kt
```diff
@@ -48,28 +49,28 @@ class AdBreakMapperTest {
...
-        assertThat(collectorAdBreak.replaceContentDuration).isEqualTo(TEN_SECONDS)
+        assertThat(collectorAdBreak.replaceContentDuration).isEqualTo(TEN_SECONDS_IN_MS)
     }
 
     @Test
     fun FromPlayerAdConfigurationWithAdConfigurationShouldNotSetValuesOfAdBreak(){
         val collectorAdBreak = adBreakMapper.FromPlayerAdConfiguration(adConfiguration)
 
-        assertThat(collectorAdBreak.scheduleTime).isEqualTo(TEN_SECONDS)
+        assertThat(collectorAdBreak.scheduleTime).isNull()
```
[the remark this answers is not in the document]
sabe25: it was a copy error
the scheduleTime prop is in the AdBreak but not in the AdConfiguration
because i feed an adConfiguration to the method there should not be a value

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -0,0 +1,236 @@
...
+        adSample.timeInViewport = 0
+        adSample.adPodPosition = this.adPodPosition
+        val timestamp = Util.getTimeStamp()
+        this.beginPlayingTimestamp = timestamp
+        // TODO this.enterViewportTimestamp = this.isContainerInViewport() ? timestamp : undefined;
+        this.isPlaying = true
+        this.currentTime = 0
+        this.adPodPosition++
+        // TODO
+//        this.currentTimeInterval = window.setInterval(() => {
```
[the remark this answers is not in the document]
sabe25: sure it is anyways the same code from web collector

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Development
 
+### Added
+
+- Added Analytics AdAdapter for Bitmovin Player
+
+### Known Issues
+
+- tracking of metrics `percentageInViewport` and `timeInView` not possible in Android
```
Tigraine:
```suggestion
- tracking of Ad `percentageInViewport` and `timeInView` is not possible in Android 
```

PR #65, opened 2019-11-18
AN-949 removed Heartbeat activation in Pause State

sabe25:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-949
- Description: during pause state the heartbeat was activated and sent out Samples
Removed the activation of hearbeat in the pause state

### Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/collector/src/main/java/com/bitmovin/analytics/stateMachines/PlayerState.java
+++ b/collector/src/main/java/com/bitmovin/analytics/stateMachines/PlayerState.java
@@ -75,7 +75,7 @@ public enum PlayerState {
                 }
             }
 
-            machine.enableHeartbeat();
+//            machine.enableHeartbeat();
 
         }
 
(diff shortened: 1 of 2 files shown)


collector/src/main/java/com/bitmovin/analytics/stateMachines/PlayerState.java
```diff
@@ -75,7 +75,7 @@ void onEnterState(PlayerStateMachine machine) {
                 }
             }
 
-            machine.enableHeartbeat();
+//            machine.enableHeartbeat();
```
Tigraine: But that would prevent the HeartBeat from running at all (it's required for when we go into the `playing` state)

sabe25: yes but the playing event does have the same actions
so the heartbeat gets disabled by the PlayExit and enabled by the PlayEntered
But with those line the heartbeat would get enabled right after it got disabled
which makes no sense to me right

PR #66, opened 2019-11-19
Feature/ad testing

rolandgriesser:
### Work
- Fixes: Correctly racking the current time in the ad

### Checklist

- [ ] Updated CHANGELOG.md (not applicable)

### DIFF
--- a/collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
+++ b/collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
@@ -4,17 +4,28 @@ import com.bitmovin.analytics.ads.*
 import com.bitmovin.analytics.data.AdSample
 import com.bitmovin.analytics.data.AdEventData
 import com.bitmovin.analytics.utils.Util
-import java.util.Collections.min
 
 class BitmovinAdAnalytics(var analytics: BitmovinAnalytics) {
     private var activeAdBreak: AdBreak? = null
     private var activeAdSample: AdSample? = null
     private var adPodPosition: Int = 0
     private var adStartupTimestamp: Long? = null
-    private var currentTime: Long? = null
+    private var _currentTime: Long? = null
     private var beginPlayingTimestamp: Long? = null
     private var isPlaying: Boolean = false
     private val adManifestDownloadTimes: HashMap<String, Long> = hashMapOf()
+
+    private val currentTime: Long?
+        get() = if(this.isPlaying) {
+            if(this._currentTime == null || this.beginPlayingTimestamp == null) {
+                null
+            } else {
+                this._currentTime!! + Util.getTimeStamp() - this.beginPlayingTimestamp!!
+            }
+
+        } else {
+            _currentTime
+        }
 
     fun onAdStarted(ad: Ad) {
         if (!ad.isLinear) {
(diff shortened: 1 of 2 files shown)


collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -4,32 +4,39 @@ import com.bitmovin.analytics.ads.*
...
         this.activeAdSample = AdSample(ad = ad)
+        this.activeAdSample!!.adStartupTime = if (this.adStartupTimestamp != null) Util.getTimeStamp() - this.adStartupTimestamp!! else null
 
-        val activeAdSample = this.activeAdSample ?: return
-
-        val adStartupTime = this.adStartupTimestamp
-        activeAdSample.adStartupTime = if (adStartupTime != null) Util.getTimeStamp() - adStartupTime else null
-
-        this.startAd(activeAdSample)
+        this.startAd(activeAdSample!!)
```
sabe25: @Tigraine @rolandgriesser would it be more secure, regarding an NullPointerException, if we assign the value of the `this.acticsAdSample` to a new variable, exit if the value is null and proceed with the local variable

like `val newAdSample = this.activeAdSample ?: return`

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -4,32 +4,39 @@ import com.bitmovin.analytics.ads.*
...
 import com.bitmovin.analytics.utils.Util
-import java.util.Collections.min
 
 class BitmovinAdAnalytics(var analytics: BitmovinAnalytics) {
     private var activeAdBreak: AdBreak? = null
     private var activeAdSample: AdSample? = null
     private var adPodPosition: Int = 0
     private var adStartupTimestamp: Long? = null
-    private var currentTime: Long? = null
+    private var _currentTime: Long? = null
```
Tigraine: What you are doing here can be achieved in pure kotlin with a backing `field`:
https://kotlinlang.org/docs/reference/properties.html#backing-fields

The `_currentTime` looks really weird

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -4,32 +4,39 @@ import com.bitmovin.analytics.ads.*
...
+        }
+
     fun onAdStarted(ad: Ad) {
         if (!ad.isLinear) {
             return
         }
 
         this.resetActiveAd()
         this.activeAdSample = AdSample(ad = ad)
+        this.activeAdSample!!.adStartupTime = if (this.adStartupTimestamp != null) Util.getTimeStamp() - this.adStartupTimestamp!! else null
```
Tigraine: agree with @sabe25  we should assign the AdSample in the end so we don't have to de-reference this here. 
(And also if we ever run in multiple threads this could be a concurrency issue if someone accesses the `this.activeAdSample` between those two assignments here

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -95,9 +98,10 @@ class BitmovinAdAnalytics(var analytics: BitmovinAnalytics) {
     }
 
     fun onPause() {
-        val activeAdSample = this.activeAdSample ?: return
         if (this.analytics.adAdapter != null && this.analytics.adAdapter.isLinearAdActive && this.activeAdSample != null) {
-            this.updatePlayingTime(activeAdSample)
+            if(this._currentTime != null && this.beginPlayingTimestamp != null) {
```
Tigraine: Should use the property w/ backing field I guess?

PR #68, opened 2019-11-27
AN-966 Fixed exoplayer-collector crash when modules have not been loaded

peterpf:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-966
- Description: Check which exoplayer modules (DASH, HLS) are loaded before checking the manifest type 

### Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/collector-exoplayer/src/main/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapter.java
+++ b/collector-exoplayer/src/main/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapter.java
@@ -23,6 +23,7 @@ import com.bitmovin.analytics.stateMachines.PlayerStateMachine;
 import com.bitmovin.analytics.utils.Util;
 import com.google.android.exoplayer2.ExoPlaybackException;
 import com.google.android.exoplayer2.ExoPlayer;
+import com.google.android.exoplayer2.ExoPlayerLibraryInfo;
 import com.google.android.exoplayer2.Format;
 import com.google.android.exoplayer2.PlaybackParameters;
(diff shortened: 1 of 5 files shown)


collector-exoplayer/src/main/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapter.java
```diff
@@ -104,6 +122,7 @@ public long getPosition() {
     @Override
     public void onTimelineChanged(Timeline timeline, Object manifest, int reason) {
         Log.d(TAG, "onTimelineChanged");
+        updateLoadedModules();
```
sabe25: would it make sense to first initialize the loaded modules in the constructor?

peterpf: Modules can be registered/unregistered anytime. 
When the `createMediaSource` is called, the respective module (HLS, DASH) is registered. At this point in time, the collector is already instantiated. So I was looking for an event that would be triggered when the manifest changes (here we can say for sure that the module was loaded because it is required). The `onTimelineChanged` event is called when the manifest changes.

rolandgriesser:
@Tigraine Opinion? I'd prefer to check if the specific class is available. Checking for the module doesn't really tell us if the class is there, so parsing could still fail.

```
public boolean isClass(String className) {
    try  {
        Class.forName(className);
        return true;
    }  catch (ClassNotFoundException e) {
        return false;
    }
}


if(isClass("com.google.android.exoplayer2.source.hls.HlsManifest"))
...
```

PR #72, opened 2019-12-06
AN-972 - Improved duration tracking

rolandgriesser:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-972
- Description: Hopefully fixes the problem with random samples having duration of around 20 years.
Initial thought was that the phone resets the time to 1.1.2000, but as the duration is positive, the phone's date is more likely in the future.
Nonetheless, this should improve how we calculate time differences, but we need to keep an eye on the data in future to see if this fixes the actual issue linked (@Tigraine)

### Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
+++ b/collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
@@ -9,17 +9,17 @@ class BitmovinAdAnalytics(var analytics: BitmovinAnalytics) {
     private var activeAdBreak: AdBreak? = null
     private var activeAdSample: AdSample? = null
     private var adPodPosition: Int = 0
-    private var adStartupTimestamp: Long? = null
-    private var beginPlayingTimestamp: Long? = null
+    private var elapsedTimeAdStartup: Long? = null
+    private var elapsedTimeBeginPlaying: Long? = null
     private var isPlaying: Boolean = false
     private val adManifestDownloadTimes: HashMap<String, Long> = hashMapOf()
 
     private var currentTime: Long? = null
         get() = if(this.isPlaying) {
-            if(field == null || this.beginPlayingTimestamp == null) {
+            if(field == null || this.elapsedTimeBeginPlaying == null) {
                 null
             } else {
-                field!! + Util.getElapsedTimestamp() - this.beginPlayingTimestamp!!
+                field!! + Util.getElapsedTime() - this.elapsedTimeBeginPlaying!!
             }
 
         } else {
@@ -33,8 +33,8 @@ class BitmovinAdAnalytics(var analytics: BitmovinAnalytics) {
 
         this.resetActiveAd()
         val adSample = AdSample(ad = ad)
-        val adStartupTimestamp = this.adStartupTimestamp
-        adSample.adStartupTime = if (adStartupTimestamp != null) Util.getElapsedTimestamp() - adStartupTimestamp else null
+        val elapsedTimeAdStartup = this.elapsedTimeAdStartup
+        adSample.adStartupTime = if (elapsedTimeAdStartup != null) Util.getElapsedTime() - elapsedTimeAdStartup else null
 
         this.activeAdSample = adSample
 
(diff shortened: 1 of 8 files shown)


collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -184,7 +184,7 @@ class BitmovinAdAnalytics(var analytics: BitmovinAnalytics) {
         eventData.setAdBreak(adBreak)
         eventData.setAdSample(adSample)
 
-        eventData.time = Util.getTimeStamp()
+        eventData.time = Util.getTimestamp()
```
sabe25: shouldn't be there `Util.getElapsedTimestamp()` ?

collector/src/main/java/com/bitmovin/analytics/data/EventData.kt
```diff
@@ -33,7 +33,7 @@ class EventData(bitmovinAnalyticsConfig: BitmovinAnalyticsConfig, val impression
     var isLive: Boolean = false
     var isCasting: Boolean = false
     var videoDuration: Long = 0
-    var time: Long = Util.getTimeStamp()
+    var time: Long = Util.getTimestamp()
```
sabe25: Same here

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
sabe25: Ok this one is the exiting payload...

collector/src/main/java/com/bitmovin/analytics/data/EventData.kt
sabe25: ok for me

collector/src/main/java/com/bitmovin/analytics/stateMachines/PlayerStateMachine.java
```diff
@@ -38,7 +38,7 @@ public PlayerStateMachine(BitmovinAnalyticsConfig config, BitmovinAnalytics anal
     public void enableHeartbeat() {
         heartbeatHandler.postDelayed(new Runnable() {
             public void run() {
-                long currentTimestamp = Util.getTimeStamp();
+                long currentTimestamp = Util.getElapsedTimestamp();
```
Tigraine: Shouldn't the naming change a bit here? `timestamp` implies to me that it's an actual Time - instead it's a random timestamp that increases monotonically that can only be used for interval calculations. 
So maybe `intervalStart` or similar.

I'm a bit worried someone might do a `sample.time = currentTimestamp` or something else

collector/src/main/java/com/bitmovin/analytics/BitmovinAdAnalytics.kt
```diff
@@ -145,8 +145,8 @@ class BitmovinAdAnalytics(var analytics: BitmovinAnalytics) {
         adSample.timePlayed = exitPosition
         adSample.playPercentage = Util.calculatePercentage(adSample.timePlayed, adSample.ad.duration, true)
 
-        // reset startupTimestamp for the next ad, in case there are multiple ads in one ad break
-        this.adStartupTimestamp = Util.getTimeStamp()
+        // reset elapseedTimeAdStartup for the next ad, in case there are multiple ads in one ad break
```
sabe25:
```suggestion
        // reset elapsedTimeAdStartup for the next ad, in case there are multiple ads in one ad break
```

collector/src/main/java/com/bitmovin/analytics/utils/Util.java
```diff
@@ -48,7 +48,7 @@ public static String getUserId(Context context) {
                 Settings.Secure.ANDROID_ID);
     }
 
-    public static long getElapsedTimestamp() {
+    public static long getElapsedTime() {
```
sabe25: would make sense to add a description here so every one knows what this method and time unit is about
```
return the time since the system was booted, and guaranteed to be monotonic
Details here: https://developer.android.com/reference/android/os/SystemClock
```

PR #73, opened 2019-12-06
Feature/improve debugging

rolandgriesser:
### Work
- Description: Added possibility to listen to debug information from the Analytics Module

@Tigraine 
Does it make sense to merge this? Otherwise just discard. It adds a `debugLog` to the ExoPlayer Sample UI and the possibility to attach a `DebugListener` to the `BitmovinAnalytics`.

### Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
+++ b/collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
@@ -15,8 +15,12 @@ import com.bitmovin.analytics.data.SimpleEventDataDispatcher;
 import com.bitmovin.analytics.stateMachines.PlayerStateMachine;
 import com.bitmovin.analytics.stateMachines.StateMachineListener;
 import com.bitmovin.analytics.license.LicenseCallback;
-import com.bitmovin.analytics.utils.DataSerializer;
 import com.bitmovin.analytics.utils.Util;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * An analytics plugin that sends video playback analytics to Bitmovin Analytics servers. Currently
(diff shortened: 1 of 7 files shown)


collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
```diff
@@ -49,15 +55,15 @@ public BitmovinAnalytics(BitmovinAnalyticsConfig bitmovinAnalyticsConfig, Contex
         this.bitmovinAnalyticsConfig = bitmovinAnalyticsConfig;
         this.playerStateMachine = new PlayerStateMachine(this.bitmovinAnalyticsConfig, this);
         this.playerStateMachine.addListener(this);
-        this.eventDataDispatcher = new SimpleEventDataDispatcher(this.bitmovinAnalyticsConfig, this.context, this);
+        this.eventDataDispatcher = new SimpleEventDataDispatcher(this.bitmovinAnalyticsConfig, this.context, this, debugCallback);
```
Tigraine: I like the idea but IMO the debugCallback should not be part of the SimpleEventDataDispatcher but rather a decorator to be more SRP

collector/src/main/java/com/bitmovin/analytics/data/DebuggingEventDataDispatcher.kt
```diff
@@ -0,0 +1,33 @@
+package com.bitmovin.analytics.data
+
+import android.content.Context
+import com.bitmovin.analytics.BitmovinAnalyticsConfig
+import com.bitmovin.analytics.DebugCallback
+import com.bitmovin.analytics.license.LicenseCallback
+
+class DebuggingEventDataDispatcher(config: BitmovinAnalyticsConfig, context: Context, callback: LicenseCallback, private val debugCallback: DebugCallback) : IEventDataDispatcher {
+
+    private val innerEventDataDispatcher = SimpleEventDataDispatcher(config, context, callback)
```
Tigraine: If we make this a constructor injected dependency then this class is stupid simple and we don't have to pass through all the constructor args for `SimpleEventDataDispatcher`

PR #77, opened 2020-03-27
feature/fix-project-dependencies

Svarozic:
### Work
- Fixes: 
- Description: Clean build.gradle files and fix bitmovinanalyticsexample build error
  - update gradle wrapper
  - define mavenCentral in root build.gradle
  - use project.ext properties to unified deps versions
  - set multiDexEnabled config only in analyticsexample app when it is needed (https://developer.android.com/studio/build/multidex#multidexkeepfile-property)

### Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/analyticsexample/build.gradle
+++ b/analyticsexample/build.gradle
@@ -3,31 +3,27 @@ apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
 
 android {
-    compileSdkVersion 27
+    compileSdkVersion project.compileSdkVersion
 
     defaultConfig {
         applicationId "com.bitmovin.exoplayeranalyticsexample"
-        minSdkVersion 16
-        targetSdkVersion 27
-        versionCode 1
-        versionName "1.0"
-        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+        minSdkVersion project.minSdkVersion
+        targetSdkVersion project.targetSdkVersion
+        versionCode project.versionCode
+        versionName project.version
 
+        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+        multiDexEnabled true
     }
 
-
     compileOptions {
-        sourceCompatibility 1.8
-        targetCompatibility 1.8
+        sourceCompatibility project.sourceCompatibility
+        targetCompatibility project.targetCompatibility
     }
 
     buildTypes {
-        debug {
-            multiDexEnabled true
-        }
         release {
             minifyEnabled false
-            multiDexEnabled true
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
         }
     }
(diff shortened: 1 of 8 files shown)


Svarozic:
I have fixed `gradle build` errors on current `develop`, because of wrong dependency to one lib from SDK 26 and I have sync deps versions in all modules with `project.ext` props

analyticsexample/build.gradle
```diff
@@ -3,31 +3,27 @@ apply plugin: 'kotlin-android'
...
     defaultConfig {
         applicationId "com.bitmovin.exoplayeranalyticsexample"
-        minSdkVersion 16
-        targetSdkVersion 27
-        versionCode 1
-        versionName "1.0"
-        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+        minSdkVersion project.minSdkVersion
+        targetSdkVersion project.targetSdkVersion
+        versionCode project.versionCode
```
rolandgriesser: Not sure if it's intended to have the collector version in the examples as well, we have never updated them before. 
@Tigraine ?

rolandgriesser:
I'm on Android Studio 3.5.2, some of the changes aren't working there anymore. Fine for me to update, but if it's not necessary, I guess we don't want to force our customers to update their dev tools, as it has been working till now. @Tigraine ?

analyticsexample/build.gradle
Tigraine: Shouldn't really matter but why not.. Makes it cleaner to debug?

collector-bitmovin-player/build.gradle
```diff
@@ -3,32 +3,27 @@ apply plugin: 'kotlin-android'
...
-
-    compileSdkVersion 27
+    compileSdkVersion project.compileSdkVersion
 
     defaultConfig {
-        minSdkVersion 16
-        targetSdkVersion 27
-        versionCode rootProject.ext.versionCode
-        versionName libraryVersion
+        minSdkVersion project.minSdkVersion
```
Tigraine: Are we sure we want to have this centralized? The library might have a different mindSdk version than the app so we can test version mismatches and stuff (e.g the customer is on 20 and our collector is built with `16` etc)

build.gradle
```diff
@@ -4,16 +4,18 @@ buildscript {
...
 
     repositories {
-        mavenLocal()
         google()
         jcenter()
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.4.2'
+        classpath 'com.android.tools.build:gradle:3.6.1'
```
rolandgriesser: Can you check if things are still working with 3.5.3? That's the newest version still supported in Android Studio 3.5. If not, we'll just update

analyticsexample/build.gradle
Svarozic: as we do not release these testing/developing apps (`bitmovinanalyticsexample` and `analyticsexample`) it should not matter, before they had always version `1.0`, now we have version in all submodules consistent.

build.gradle
Svarozic: I do not know, how to downgrade Android Studio plugin to 3.5 to test it :( but this change was probably done because of some warning/note popup from latest AndroidStudio, so that is why I have changed it ... could you please update you IDE 

> I guess we don't want to force our customers to update their dev tools, as it has been working till now. @Tigraine ?

This change affects only us developers, customers used built versions of our collector library

build.gradle
```diff
@@ -22,23 +24,39 @@ allprojects {
...
+    targetSdkVersion = 27
+
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+
+    constraintLayoutVersion = '1.0.2'
+    junitVersion = '4.12'
+    testRunnerVersion = '1.0.1'
+    espressoCoreVersion = '3.0.1'
+    playercoreVersion = '2.35.0'
```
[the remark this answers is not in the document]
Svarozic: I did not want to "rename" lib that we use
```
implementation "com.bitmovin.player:playercore:2.35.0"
```

but at least I can rename it to `bitmovinPlayerCoreVersion`

collector-bitmovin-player/build.gradle
Svarozic: we should have it centralised, as we want to have 3x modules (`collector`, `collector-bitmovin-player`, `collector-exoplayer`) written in way to support this minSDK 16.

If it is needed and we want to test version mismatches, we can change/increase minSdk version of our testing/development apps (`analyticsexample`, `bitmovinanalyticsexample`) to latest SDK version. I have not done this as I wanted to stay with minSdk version we had before, and that was `v16` in all modules.

build.gradle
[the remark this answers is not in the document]
rolandgriesser: Yes makes sense. Would prefer `bitmovinPlayerCoreVersion` as otherwise in the root gradle file it's not obvious which player we're talking about.

build.gradle
rolandgriesser: Ah so in Android Studio >= 3.6, old gradle versions aren't working? Didn't know that, will update then :)

Svarozic: let me know once you update Android Studio (and plugins, as it is necessary as well, most of the time when IntelliJ is updated) how it works by you,  I will wait with merge for you approve.

PR #78, opened 2020-03-30
AN-1236 Add volume change tracking

KulenovicInes:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1236
- Description: Added content listener for volume changes and util method for getting current volume level.

### Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
+++ b/collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
@@ -3,6 +3,7 @@ package com.bitmovin.analytics;
 import static com.bitmovin.analytics.utils.DataSerializer.serialize;
 
 import android.content.Context;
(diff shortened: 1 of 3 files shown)


collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
```diff
@@ -278,11 +292,22 @@ public void onAudioTrackChange() {
     EventData data = playerAdapter.createEventData();
     data.setState(playerStateMachine.getCurrentState().toString().toLowerCase());
     data.setDuration(0);
+    data.setVolume(Util.getDeviceVolume(context));
```
sabe25: Note:
Instead of adding the `setVolume` to every Event you could put this into the `createEventData` of the adapters

PR #82, opened 2020-04-10
AN-1083 track playback start failure

KulenovicInes:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1083
- Description: Added play attempt metric on android bitmovin player collector

### Checklist

- [ ] Updated CHANGELOG.md


collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
```diff
@@ -184,8 +186,14 @@ public void onError(ErrorCode errorCode) {
         Log.d(TAG, String.format("onError %s", playerStateMachine.getImpressionId()));
         EventData data = playerAdapter.createEventData();
         data.setState(playerStateMachine.getCurrentState().toString().toLowerCase());
+
+        if (playerStateMachine.getCurrentState() == PlayerState.SETUP) {
+            data.setVideoStartFailed(true);
+            data.setVideostartFailedReason(VideoStartFailedReason.PLAYER_ERROR.toString().toLowerCase());
+        }
```
sabe25, line 193: Do we really track this as a play attempt?
Does it count as a play attempt if the player errors during setup?

collector/src/main/java/com/bitmovin/analytics/enums/VideoStartFailedReason.java
```diff
@@ -0,0 +1,8 @@
+package com.bitmovin.analytics.enums;
+
+public enum VideoStartFailedReason {
+    PAGE_CLOSED,
+    PLAYER_ERROR,
+    TIMEOUT,
+    UNKNOWN
+}
```
sabe25, line 8: make this a Kotlin class and try something like this
```
enum class VideoStartFailedReason(val reason: String) {
    PAGE_CLOSE("page_close")
}
```
This way it is much easier to access the string value of the enum `VideoStartFailedReason.PAGE_CLOSE.reason`

collector/src/main/java/com/bitmovin/analytics/data/EventData.kt
```diff
@@ -73,4 +73,6 @@ class EventData(bitmovinAnalyticsConfig: BitmovinAnalyticsConfig, val impression
     var audioLanguage: String? = null
     var drmType: String? = null
     var drmLoadTime: Long? = null
+    var videoStartFailed: Boolean? = false
+    var videostartFailedReason: String? = null
```
sabe25, line 77:
```suggestion
    var videoStartFailedReason: String? = null
```

collector/src/main/java/com/bitmovin/analytics/data/EventData.kt
```diff
@@ -73,4 +73,6 @@ class EventData(bitmovinAnalyticsConfig: BitmovinAnalyticsConfig, val impression
     var audioLanguage: String? = null
     var drmType: String? = null
     var drmLoadTime: Long? = null
+    var videoStartFailed: Boolean? = false
```
sabe25, line 76: no default value needed
```suggestion
    var videoStartFailed: Boolean?
```

collector/src/main/java/com/bitmovin/analytics/stateMachines/PlayerStateMachine.java
```diff
@@ -27,6 +28,7 @@
     private Handler heartbeatHandler = new Handler();
     private int heartbeatDelay = 59700; // default to 60 seconds
     private final BitmovinAnalytics analytics;
+    private VideoStartFailedReason videoStartFailedReason;
```
sabe25, line 31: can we move this away from the State Machine?
Because the state machine does not intentionally work with the values of the state
for me this is more of a thing for the adapter

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
```diff
@@ -289,6 +297,21 @@ public void onAudioTrackChange() {
...
     }
 
+    @Override
+    public void onVideoStartFailed() {
+        String videoStartFailedReason = playerStateMachine.getVideoStartFailedReason() != null
+                ? playerStateMachine.getVideoStartFailedReason().toString().toLowerCase()
+                : VideoStartFailedReason.UNKNOWN.toString().toLowerCase();
+
+        EventData data = playerAdapter.createEventData();
+        data.setState(playerStateMachine.getCurrentState().toString().toLowerCase());
```
sabe25, line 307: do we want to set the state to `EXITBEFOREVIDEOSTART`?
i noticed in the web collector that this never happend

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
KulenovicInes: If e.g. the first segment is corrupted then error will occur after user pressed play and player will never go into playing state. I'm not sure if there is some other case when error may happen after pressing play and actual start of playing. @peterpf what do you think does this case counts as play attempt?

collector/src/main/java/com/bitmovin/analytics/stateMachines/PlayerStateMachine.java
KulenovicInes: We have multiple possible reasons why play attempt failed (timeout, some error...), and this reason is determined in some of the listeners in`BitmovinSdkAdapter`. We need some way of passing that information to the `BitmovinAnalytics`.

One option was to use similar logic as it now in case of onError event (`ErrorCode class`), but then this class would need some changes, and I'm not sure if it makes sense to use it for this case.

I'm open for any suggestions/improvements for this case.

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
KulenovicInes: I think web collector always sends some kind of state if failed attempt happens, and usually it's startup state. This field is nullable so we don't need to set it. Just wanted to make sure we are consistent with value of states.

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
```diff
@@ -473,8 +502,12 @@ public void onError(ErrorEvent errorEvent) {
                     break;
             }
             stateMachine.setErrorCode(errorCode);
-            stateMachine.transitionState(PlayerState.ERROR, getPosition());
-
+            if (!isVideoPlayed) {
+                stateMachine.setVideoStartFailedReason(VideoStartFailedReason.PLAYER_ERROR);
+                stateMachine.transitionState(PlayerState.EXITBEFOREVIDEOSTART, getPosition());
+            } else {
+                stateMachine.transitionState(PlayerState.ERROR, getPosition());
```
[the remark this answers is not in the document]
KulenovicInes: We can transition to ERROR state, but then we must have some value of `ErrorCode`. 
If I figured out everything correctly, similarly to ERROR state logic, in `transitionState` method before actual state is set, `onEnterState` is called which then calls other listener methods. So in moment of executing listener method, change in actual state should not be present, and there should be value of previous state (setup).

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
peterpf: PlayAttempts should be only tracked after the `SETUP` finished. Errors before that are just normal errors, no playAttempt failures.

collector/src/main/java/com/bitmovin/analytics/data/EventData.kt
KulenovicInes: That way there is `Property must be initialized or be abstract` error.

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
sabe25: Ahh got it... it sets the State on enterState and then it contains the old state 👍

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
```diff
@@ -264,13 +272,19 @@ public DRMInformation getDRMInformation() {
...
         }
     };
 
     private OnSourceUnloadedListener onSourceUnloadedListener = new OnSourceUnloadedListener() {
         @Override
         public void onSourceUnloaded(SourceUnloadedEvent sourceUnloadedEvent) {
             Log.d(TAG, "On Source Unloaded");
+            if (!isVideoPlayed && isVideoAttemptedPlay) {
+                stateMachine.setVideoStartFailedReason(VideoStartFailedReason.PAGE_CLOSED);
+                stateMachine.transitionState(PlayerState.EXITBEFOREVIDEOSTART, getPosition());
```
peterpf, line 286: When an error happens the player is also unloading the source. Thus, there will be two samples sent out with the videoStartFailed data. Another reason to not use this  method is because when `player.load(...)` is called, the `unload` method will be also invoked. Please try to use a dedicated event to detect when the app closes, e.g. the `onDestroy` (not sure if this works 100% of the time): https://developer.android.com/reference/android/app/Activity#ActivityLifecycle

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
```diff
@@ -509,10 +541,25 @@ public void onReady(ReadyEvent readyEvent) {
...
     };
+
+    private CountDownTimer videoStartTimeout = new CountDownTimer(Util.VIDEOSTART_TIMEOUT, 1000) {
+        @Override
+        public void onTick(long millisUntilFinished) {
+        }
+
+        @Override
+        public void onFinish() {
+            Log.d(TAG, "VideoStartTimeout finish");
```
peterpf, line 560: @sabe25 In the iOS collector the timer stops when the app is put into background, right?

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
```diff
@@ -509,10 +541,25 @@ public void onReady(ReadyEvent readyEvent) {
 
             if (bitmovinPlayer.isPlaying()) {
                 stateMachine.transitionState(PlayerState.PLAYING, getPosition());
+                // autoplay
+                isVideoAttemptedPlay = true;
```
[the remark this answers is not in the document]
KulenovicInes: I wasn't sure where to put it, since in this case we don't have any onPlay event. I can start time here in onReady maybe?

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
KulenovicInes: This `onDestry()` event is from entire app, so I'm not really sure how to pass this event up to `BitmovinAnalytics` (this might be possible by adding new method here) and up to the `BitmovinSdkAdapter` where are all the flags and timer.

I can see that player has it internal `onDestroyedListener` so maybe it can be used for case when app is closed?

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
sabe25: Yes it pauses the timer when it goes into background and continue when it comes back into foreground/focus
For android there is the onPause event
https://developer.android.com/reference/android/app/Activity#onPause()

KulenovicInes: I think it's the same thing as with onDestroy event. These events are related to the main app, and track only if entire app goes into specific state. I could maybe add some new method logic in `BitmovinAnalytics` file, but then it would depend on calls form "outside" to stop timer. This would also might require moving timer logic from `BitmovinSdkAdapter` to `BitmovinAnalytics` and I'm not sure if that is good option because `BitmovinAnalytics` should be only responsible for crating and sending event data (at least as I see it).

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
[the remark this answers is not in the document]
KulenovicInes: Added start of timer here onReady

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
KulenovicInes: Added onDestroy event listener

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
peterpf: I see. Thanks for elaborating the problem.

PR #85, opened 2020-04-22
AN-1124 data downloaded metrics for ExoPlayer

KulenovicInes:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1124
- Description: Implementation is based on one from web collector. Fields exist on ingress app and some of them are commented out, but AFAIK those fields don't exist in CrateDB. Field types are based on field types in ingress, so there should be no issue in deserializing them. Some fields are not used at the moment (like httpStatus and timeToFirstByte) but they should be relevant for this implementation with Bitmovin player.

### Checklist

- [ ] Updated CHANGELOG.md


collector-exoplayer/src/main/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapter.java
```diff
@@ -379,6 +385,16 @@ else if (mediaLoadData.dataType == DATA_TYPE_MEDIA &&
             this.drmType = drmType;
         }
 
+        if(mediaLoadData.trackFormat != null
```
Tigraine, line 388: Small nitpick but i feel like the onLoadCompleted callback is already doing too much here.
Might make sense to refactor it a bit and break it into different methods that are named after the different things that are happening (manifest, drm, videosegment)

Tigraine:
Good to go 👍

PR #86, opened 2020-04-27
AN-1281 fixed rebuffering after seek in exoplayer

sabe25:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1281
- Description: seek start event was listening in the wrong method

### Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/collector-exoplayer/src/main/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapter.java
+++ b/collector-exoplayer/src/main/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapter.java
@@ -169,7 +169,7 @@ public class ExoPlayerAdapter implements PlayerAdapter, Player.EventListener, An
                 }
                 break;
             case Player.STATE_BUFFERING:
-                if (this.stateMachine.getCurrentState() != PlayerState.SEEKING && this.stateMachine.getElapsedTimeFirstReady() != 0) {
+                if (this.stateMachine.getElapsedTimeFirstReady() != 0) {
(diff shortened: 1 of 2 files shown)


collector-exoplayer/src/main/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapter.java
```diff
@@ -169,7 +169,7 @@ public void onPlayerStateChanged(boolean playWhenReady, int playbackState) {
                 }
                 break;
             case Player.STATE_BUFFERING:
-                if (this.stateMachine.getCurrentState() != PlayerState.SEEKING && this.stateMachine.getElapsedTimeFirstReady() != 0) {
+                if (this.stateMachine.getElapsedTimeFirstReady() != 0) {
```
rolandgriesser, line 172: Are we sure we can just remove that? Otherwise the PR looks good to me

sabe25: i was also thinking that this was there on purpose
Testet it very well to understand it, but removed it in the end
Otherwise there are no rebuffering events after seeking
This might come from bitmovin player but here it works a little bit different

Bitmovin player
After seek has finished the player automatically transitions back to play
if no content ready -> rebuffering

Exo
Here we don't transition to play 
so we have a direct transition from seek to rebuffering

rolandgriesser: If you tested it, then we're good to go 👍

PR #91, opened 2020-04-30
AN-1349 android tv check

sabe25:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1349
- Description: added field to check if device is a TV
https://developer.android.com/training/tv/start/hardware.html#java

### Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/collector/src/main/java/com/bitmovin/analytics/data/DeviceInformationProvider.kt
+++ b/collector/src/main/java/com/bitmovin/analytics/data/DeviceInformationProvider.kt
@@ -1,10 +1,17 @@
 package com.bitmovin.analytics.data
 
+import android.app.UiModeManager
 import android.content.Context
+import android.content.res.Configuration
 import android.os.Build
 import com.bitmovin.analytics.utils.Util
 
-open class DeviceInformationProvider(val context: Context, val userAgent: String, val isTV: Boolean) {
+open class DeviceInformationProvider(val context: Context, val userAgent: String) {
+    private var isTV: Boolean = false
+    init {
+        val uiModeManager = context.getSystemService(Context.UI_MODE_SERVICE) as UiModeManager
+        isTV = uiModeManager.currentModeType == Configuration.UI_MODE_TYPE_TELEVISION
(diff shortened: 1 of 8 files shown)


collector/src/main/java/com/bitmovin/analytics/data/DeviceInformationProvider.kt
```diff
@@ -4,11 +4,12 @@ import android.content.Context
 import android.os.Build
 import com.bitmovin.analytics.utils.Util
 
-open class DeviceInformationProvider(val context: Context, val userAgent: String) {
+open class DeviceInformationProvider(val context: Context, val userAgent: String, val isTV: Boolean) {
     fun getDeviceInformation(): DeviceInformation {
         return DeviceInformation(
                 manufacturer = Build.MANUFACTURER,
                 model = Build.MODEL,
+                isTV = isTV,
```
Tigraine, line 12: I think it makes more sense to just use the `context` and extract the info here :)
The responsibility / name of this class is `DeviceInformationProvider`. It's actually a Factory for DeviceInformation so the logic "is this a tv" would totally fit here and would also be easily tested.
Also the benefit would be that the call site in `BitmovinSdkAdapter` doesn't have to change :)

With this you are actually "spreading" the responsibility for the `isTV` to multiple places

collector/src/main/java/com/bitmovin/analytics/data/DeviceInformationProvider.kt
```diff
@@ -1,14 +1,22 @@
...
 
+import android.app.UiModeManager
 import android.content.Context
+import android.content.res.Configuration
 import android.os.Build
 import com.bitmovin.analytics.utils.Util
 
 open class DeviceInformationProvider(val context: Context, val userAgent: String) {
+    private var isTV: Boolean = false
+    init {
```
Tigraine, line 11: nothing wrong with having this in a function and calling it from `getDeviceInformation` ..

PR #94, opened 2020-05-05
AN-1409 exoplayer start failure

KulenovicInes:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1409
- Description: Updated version of exoplayer and added timeout for start playing video.

### Checklist

- [X] Updated CHANGELOG.md


KulenovicInes:
@sabe25 to be able to track if player went to background (or if it's destroy)it would be necessary to implement something like `LifecycleObserver` on main app, because it not possible other way to know what is happening on app from the collector side. This would be breaking change because it might (probably) require change on client side of the way analytics collector is initialized.

collector-exoplayer/src/main/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapter.java
```diff
@@ -306,12 +326,43 @@ private void decorateDataWithPlaybackInformation(EventData data) {
...
 
+    @Override
+    public void onPlaybackSuppressionReasonChanged(int playbackSuppressionReason) {
+        Log.d(TAG, "onPlaybackSuppressionReasonChanged " + playbackSuppressionReason);
+    }
+
+    @Override
+    public void onIsPlayingChanged(boolean isPlaying) {
+
+    }
```
sabe25, line 338: Can we remove those duplicate methods?

KulenovicInes:
> only on little request then its G2G
> 
> Where do we set the videofailed boolean from the sample?
> I guess it's in the statemachine and already mergen with the bitmovin player right?

yes, this part was implemented with changes on bitmovin player

collector-exoplayer/src/main/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapter.java
KulenovicInes: I checked this methods. They have same name, same number of parameters (and almost same names of parameters), but same parameters are of different type. Those methods come from event listener, and analytics listener interfaces.

PR #101, opened 2020-05-19
AN-1456 added error data

peterpf:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1456
- Description:
  - Adapted `ExceptionMapper` interface
  - Moved creation of error data object into class `BitmovinPlayerExceptionMapper` 
  - Check if `data` from the player errorEvent is an exception and add stacktrace to errorData

### Checklist

- [x] Updated CHANGELOG.md


collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinPlayerExceptionMapper.kt
```diff
@@ -0,0 +1,94 @@
...
+            else -> {
+                errorCode = ErrorCode.UNKNOWN_ERROR
+                errorCode.description = event.message
+            }
+        }
+        if (event.data is Throwable) {
+            val exception = event.data as Throwable
+
+            errorCode.errorData = ErrorData(exception.message
+                    ?: "", exception.topOfStacktrace)
```
KulenovicInes, line 90: Only thing here, are we sure this will not be some really long message?

peterpf: The `topOfStackTrace` method only returns a specific number of rows :)

collector/src/main/java/com/bitmovin/analytics/error/ExceptionMapper.kt
```diff
@@ -2,6 +2,6 @@ package com.bitmovin.analytics.error
 
 import com.bitmovin.analytics.data.ErrorCode
 
-interface ExceptionMapper<in T : Throwable> {
+interface ExceptionMapper<in T> {
```
Tigraine, line 5: Why did you remove this here? You should probably add the ExoPlayer subclass of ExeptionMapper have the `T : Throawable` since you removed it to be able to extend the bitmovin player here

peterpf: Removed to be able to pass `ExoplayerPlabackException` and `ErrorEvent` as type. The `ExoPlayerExceptionMapper` already has `Throwable` as type.

PR #111, opened 2020-06-22
feature/AN-1577_set-drmLoadTime-only-in-first-row

Svarozic:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1577
- Description: 

### Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Development
 
+### Changed
+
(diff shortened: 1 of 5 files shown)


Svarozic:
> Hmm. 🤔
> Does this set the `drmType` across all samples?

yes, `collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java` sets both (type and **loadTime**) only in `startup` handler, but `type` is set all the time in each adapter in `createEventData` method, when base sample is created (for other handlers) and later filled with other sample data

PR #113, opened 2020-07-16
AN-1625 AN-1619 fixing startup workflow

sabe25:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1619
  - Also Fixes: https://bitmovin.atlassian.net/browse/AN-1625
- Description: 

### Checklist

- [ ] Updated CHANGELOG.md


collector/src/main/java/com/bitmovin/analytics/stateMachines/PlayerStateMachine.java
```diff
@@ -173,6 +182,32 @@ public void setVideoStartFailedReason(VideoStartFailedReason videoStartFailedRea
         this.videoStartFailedReason = videoStartFailedReason;
     }
 
+    public CountDownTimer videoStartTimeout = new CountDownTimer(Util.VIDEOSTART_TIMEOUT, 1000) {
+        @Override
+        public void onTick(long millisUntilFinished) {
```
KulenovicInes, line 187: Does it maybe make sense to make this method protected and move entire timer logic to the state machine? AFAICS already everything is happening inside `PlayerState`, so only change would be to move stopping timer `onError`?

KulenovicInes:
Looks good to me. I tried to run both apps with regular and corrupted sources, with and without autoplay, bitmovin player with and without ads, tried to change source.... And I didn't see something breaking or not functioning.
Checked samples in DB and they also look ok to me...

If there is some edge case to pay attention I can give it another try, but otherwise looks good to go.

collector/src/main/java/com/bitmovin/analytics/stateMachines/PlayerStateMachine.java
sabe25: Thanks for this nice catch 👍 
https://github.com/bitmovin/bitmovin-analytics-collector-android/pull/113/commits/cbc59fbeb7730c124364265743594df9cb513459

collector/src/main/java/com/bitmovin/analytics/stateMachines/PlayerStateMachine.java
```diff
@@ -130,7 +135,11 @@ private void setCurrentState(final PlayerState newPlayerState) {
     }
 
     public long getStartupTime() {
-        return elapsedTimeFirstReady - elaspedTimeInitial;
+        return startupTime;
+    }
+
+    public void addStartupTime(long elapsedTime) {
+        this.startupTime =+ elapsedTime;
```
peterpf, line 142: interesting 🤔

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
```diff
@@ -454,42 +477,17 @@ public void onError(ErrorEvent errorEvent) {
...
-        }
-
+    private OnAdBreakStartedListener onAdBreakStartedListener = new OnAdBreakStartedListener() {
         @Override
-        public void onFinish() {
-            Log.d(TAG, "VideoStartTimeout finish");
-            stateMachine.setVideoStartFailedReason(VideoStartFailedReason.TIMEOUT);
-            stateMachine.transitionState(PlayerState.EXITBEFOREVIDEOSTART, getPosition());
+        public void onAdBreakStarted(AdBreakStartedEvent adBreakStartedEvent) {
+            stateMachine.startAd(getPosition());
```
rolandgriesser, line 490: How do you get out of the ad state again? Looks like you track all the play/pause etc during an ad as well.

sabe25: When the ad has finished there is a state transition flow like this
Ad playing
onPlay
AD -> Startup
onPlaying
Startup -> Playing

Do you see issues here?
No Play event listening during ads?

PR #116, opened 2020-07-24
AN-1560 restrict quality changes

KulenovicInes:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1560
- Description: Added counter for tracking number of quality changes and timer for resetting counter

### Checklist

- [X] Updated CHANGELOG.md

### DIFF
--- a/collector-exoplayer/src/test/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapterTest.kt
+++ b/collector-exoplayer/src/test/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapterTest.kt
@@ -35,11 +35,24 @@ class ExoPlayerAdapterTest {
     fun testNoStateTransitionToQualityChangeIfBitrateDidNotChange() {
         val bitrate = 3000
         `when`(stateMachine.currentState).thenReturn(PlayerState.PLAYING)
+        `when`(stateMachine.isQualityChangeEventEnabled).thenReturn(true)
         adapter.fakePosition = 20
         adapter.onDecoderInputFormatChanged(getEventTime(20L), 0, Format.createVideoSampleFormat(null, null, null, bitrate, 1, 300, 300, 64F, null, null))
         verify(stateMachine, times(1)).transitionState(eq(PlayerState.QUALITYCHANGE), ArgumentMatchers.anyLong())
+
         adapter.onDecoderInputFormatChanged(getEventTime(30L), 0, Format.createVideoSampleFormat(null, null, null, bitrate, 1, 300, 300, 64F, null, null))
         verify(stateMachine, times(1)).transitionState(eq(PlayerState.QUALITYCHANGE), ArgumentMatchers.anyLong())
+    }
+
+    @Test
+    fun testNoStateTransitionToQualityChangeIfQualityChangeLimit() {
+        val bitrate = 3000
+        `when`(stateMachine.currentState).thenReturn(PlayerState.PLAYING)
+        `when`(stateMachine.isQualityChangeEventEnabled).thenReturn(false)
+        adapter.fakePosition = 20
+        adapter.onDecoderInputFormatChanged(getEventTime(20L), 0, Format.createVideoSampleFormat(null, null, null, bitrate, 1, 300, 300, 64F, null, null))
+        verify(stateMachine, times(0)).transitionState(eq(PlayerState.QUALITYCHANGE), ArgumentMatchers.anyLong())
+
     }
 
     private fun getEventTime(realTime: Long): AnalyticsListener.EventTime {
--- a/collector-exoplayer/src/test/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapterTest.kt
+++ b/collector-exoplayer/src/test/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapterTest.kt
@@ -52,7 +52,6 @@ class ExoPlayerAdapterTest {
         adapter.fakePosition = 20
         adapter.onDecoderInputFormatChanged(getEventTime(20L), 0, Format.createVideoSampleFormat(null, null, null, bitrate, 1, 300, 300, 64F, null, null))
         verify(stateMachine, times(0)).transitionState(eq(PlayerState.QUALITYCHANGE), ArgumentMatchers.anyLong())
-
     }
 
     private fun getEventTime(realTime: Long): AnalyticsListener.EventTime {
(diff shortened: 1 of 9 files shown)


collector-exoplayer/src/test/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapterTest.kt
```diff
@@ -35,13 +35,25 @@ class ExoPlayerAdapterTest {
     fun testNoStateTransitionToQualityChangeIfBitrateDidNotChange() {
         val bitrate = 3000
         `when`(stateMachine.currentState).thenReturn(PlayerState.PLAYING)
+        `when`(stateMachine.isQualityChangeEventEnabled).thenReturn(true)
         adapter.fakePosition = 20
         adapter.onDecoderInputFormatChanged(getEventTime(20L), 0, Format.createVideoSampleFormat(null, null, null, bitrate, 1, 300, 300, 64F, null, null))
         verify(stateMachine, times(1)).transitionState(eq(PlayerState.QUALITYCHANGE), ArgumentMatchers.anyLong())
+
         adapter.onDecoderInputFormatChanged(getEventTime(30L), 0, Format.createVideoSampleFormat(null, null, null, bitrate, 1, 300, 300, 64F, null, null))
         verify(stateMachine, times(1)).transitionState(eq(PlayerState.QUALITYCHANGE), ArgumentMatchers.anyLong())
```
peterpf, line 44:
```suggestion
        verify(stateMachine, times(0)).transitionState(eq(PlayerState.QUALITYCHANGE), ArgumentMatchers.anyLong())
```
The name of the test suggests that we do not expect a second quality change event, is this intended?

collector/src/main/java/com/bitmovin/analytics/stateMachines/PlayerState.java
```diff
@@ -124,13 +125,27 @@ void onExitState(PlayerStateMachine machine, long elapsedTime, PlayerState desin
...
-            for (StateMachineListener listener : machine.getListeners()) {
-                listener.onQualityChange();
-            }
+           if(machine.isQualityChangeEventEnabled()) {
+               for (StateMachineListener listener : machine.getListeners()) {
+                   listener.onQualityChange();
+               }
+           }
+           else {
+               ErrorCode errorCode = new ErrorCode(10000, "ANALYTICS_QUALITY_CHANGE_THRESHOLD_EXCEEDED", null );
```
peterpf, line 142: Please create a constant for the message and error-code. I think we should already have something in place for the buffering -- if not, we will need it anyways.

collector/src/main/java/com/bitmovin/analytics/utils/Util.java
```diff
@@ -23,6 +23,8 @@
     public static final String SMOOTH_STREAM_FORMAT = "smooth";
     public static final int MILLISECONDS_IN_SECONDS = 1000;
     public static final int VIDEOSTART_TIMEOUT = 1000 * 60; // in milliseconds
+    public static final int ANALYTICS_QUALITY_CHANGE_COUNT_THRESHOLD = 50;
+    public static final int ANALYTICS_QUALITY_CHANGE_COUNT_RESET_INTERVAL = 60 * 60 * 1000;
```
peterpf, line 27: Please add a comment which unit is used (milliseconds, seconds). Makes it clearer later on what this should be.

collector/src/main/java/com/bitmovin/analytics/stateMachines/PlayerState.java
KulenovicInes: Created `AnalyticsErrorCodes` enum for quality change and buffering timeout errors, please check it out. This same class will also be in my PR for limiting duration of buffering event.

collector-exoplayer/src/test/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapterTest.kt
KulenovicInes: Hmm, `testNoStateTransitionToQualityChangeIfBitrateDidNotChange` works as it did previously... I only added mock for additional condition `isQualityChangeEventEnabled`. 
As far as I can see first call to state machine should be when `onDecoderInputFormatChanged` event is first time triggered (on line 40). Then on line 43 we trigger it again, with same parameters. Now because there has been no change to bitrate (parameters of both events are the same) state machine is not called again. I think that is why on line 44 we have check to confirm that only one call has been made. 🤔

collector/src/main/java/com/bitmovin/analytics/stateMachines/PlayerStateMachine.java
```diff
@@ -221,6 +225,33 @@ public void startAd(long position){
...
+
+    public void increaseQualityChangeCount(){
+        this.qualityChangeCount++;
+    }
+
+    protected void resetQualityChangeCount(){
+        this.qualityChangeCount = 0;
+    }
+
+    protected CountDownTimer qualityChangeResetTimeout = new CountDownTimer(Util.ANALYTICS_QUALITY_CHANGE_COUNT_RESET_INTERVAL, 1000) {
```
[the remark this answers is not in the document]
KulenovicInes: I checked his PR, but as far as I can see he added stopping of timer in `release()` methods on Bitmovin and Exoplayer adapters... Then our refactoring happened and this timer has became `protected` and this cancel method is now in `resetStateMachine()` method of `PlayerStateMachine` class. Is this the correct way? Should we change it back to `public` and cancel it in `release()` methods of adapters?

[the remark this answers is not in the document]
rolandgriesser: oh, when did that happen? so we refactored his memory leak fix? 
if `resetStateMachine` is called in `release`, then we're fine and it's enough doing it there, if not, we have to call both `       videoStartTimeout.cancel();` and this `qualityChangeResetTimeout.cancel()` in release (both Exo and BM).
Or even better, have a `dispose` or `release` method in the statemachine itself. (or as said, probably just calling `resetStateMachine` would be enough as everything is happening there anyways

collector-exoplayer/src/test/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapterTest.kt
peterpf: Ah got it. Mistakenly though that the `verify(..., times(N))` will be reset every time this is called 🤦. Thanks for clarifying :)

collector/src/main/java/com/bitmovin/analytics/stateMachines/PlayerStateMachine.java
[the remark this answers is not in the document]
KulenovicInes: Added `stateMachine.resetStateMachine();` on `release` methods in both collectors. Like you said this method already has logic for disabling heartbeats and looks like a good place to cancel timer (beside `videoStartFailure` we will now have `qualityChange` and `rebufferTime` timers). Can you please check it out?

PR #117, opened 2020-07-24
AN-1559 threshold buffering

KulenovicInes:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1559
- Description: Added timeout for rebuffering event.

`AnalyticsErrorCodes` class and canceling of rebufferi timer is the same as in restrict quality changes PR (https://github.com/bitmovin/bitmovin-analytics-collector-android/pull/116)

### Checklist

- [X] Updated CHANGELOG.md

### DIFF
--- a/collector/src/main/java/com/bitmovin/analytics/stateMachines/PlayerStateMachine.java
+++ b/collector/src/main/java/com/bitmovin/analytics/stateMachines/PlayerStateMachine.java
@@ -7,6 +7,7 @@ import android.util.Log;
 import com.bitmovin.analytics.BitmovinAnalytics;
 import com.bitmovin.analytics.BitmovinAnalyticsConfig;
 import com.bitmovin.analytics.data.ErrorCode;
+import com.bitmovin.analytics.enums.AnalyticsErrorCodes;
 import com.bitmovin.analytics.enums.VideoStartFailedReason;
 import com.bitmovin.analytics.utils.Util;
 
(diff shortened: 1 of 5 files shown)


collector/src/main/java/com/bitmovin/analytics/stateMachines/PlayerStateMachine.java
```diff
@@ -221,6 +223,22 @@ public void startAd(long position){
...
+        @Override
+        public void onTick(long millisUntilFinished) {
+        }
+
+        @Override
+        public void onFinish() {
+            Log.d(TAG, "rebufferingTimeout finish");
+            setErrorCode(AnalyticsErrorCodes.ANALYTICS_BUFFERING_TIMEOUT_REACHED.getErrorCode());
+            transitionState(PlayerState.ERROR, analytics.getPosition());
+            disableRebufferHeartbeat();
```
rolandgriesser, line 237: On Web, we only transition out of the error state with a source change (also the intended behaviour here, as we don't want to go back to playing/seeking/pausing etc once we run into this error). I'd call `resetStatemachine` here, so everything gets disabled and we are back at `READY` state. And then, in the `isTransitionAllowed`, I'd add the same code like on iOS (I think this covers everything that we need for now, might be extended in future OR even better, replaced by the new statemachine :D ):
```
} else if state == .ready && (destinationState != .error && destinationState != .playAttemptFailed && destinationState != .startup && destinationState != .ad) {
            return false
        }
```

KulenovicInes: Added suggested transition check so now from `ready` state we can go only to `error`, `exitbeforevideostart`, `startup` and `ad`.

Also added reset of state machine on timer end. This means that if someone waits for 2 minutes of buffering and miraculously stream comes to life, collector will emit `startup` and `playing` events with new `impressionId` and continue working normally. I don't think this is realistic use case tho :) 
(really looking forward to new `statemachine` :) )

PR #118, opened 2020-08-04
AN-1506 device available width and height

KulenovicInes:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1506
- Description: 

https://developer.android.com/reference/android/view/WindowManager#getDefaultDisplay()
https://developer.android.com/reference/android/view/Display#getMetrics(android.util.DisplayMetrics)

To be able to use `getRealMetrics()` we would need to update minSdkVersion to 17 (android 4.2)

https://developer.android.com/reference/android/view/Display#getRealMetrics(android.util.DisplayMetrics)

In Jira ticket there is a updated table with dimensions that were sent previously, values these methods return and with results of calculations.

As discussed with Roland, we need to include dimensions of status bar, so best approach would be to use real dimensions and divide them with screen density. As can be seen in table, values we get this way are similar to those we would get using `context.resources.configuration`, since that method returns available dimensions without status bar or some other screen decoration elements.

### Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Development
 
 ### Changed
-- Way of getting available screen width and height
+- Calculations for available screen width and height to now include screen density and be more independent of physical pixel
 
 ## v1.18.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,15 +7,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Development
 
+### Changed
+- Way of getting available screen width and height
+
 ## v1.18.0
 
 ### Added
 
 - Restricted number of quality changes tracked per period of time (AN-1560)
 - Timeout for rebuffering event (AN-1559)
-
-### Changed
-- Getting available screen dimensions
 
 ## v1.17.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Restricted number of quality changes tracked per period of time (AN-1560)
 - Timeout for rebuffering event (AN-1559)
 
+### Changed
+- Getting available screen dimensions
+
 ## v1.17.0
 
 ### Changed
--- a/collector/src/main/java/com/bitmovin/analytics/data/DeviceInformationProvider.kt
+++ b/collector/src/main/java/com/bitmovin/analytics/data/DeviceInformationProvider.kt
@@ -3,16 +3,14 @@ package com.bitmovin.analytics.data
 import android.app.UiModeManager
 import android.content.Context
 import android.content.res.Configuration
-
 import android.os.Build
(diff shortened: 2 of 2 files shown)


KulenovicInes:
> So this now gives us the actual "density independent" pixels? I thought that's what the "dp" stuff from Android was about anyways? Why don't we use `screenWidthDp `?

Because eg. `screenHeightDp` does not include status bar height. According to documentation values form those methods don't include "screen decoration elements".  When discussing this with @rolandgriesser we concluded that it would maybe be more accurate measurement if we used real dimensions. 🤔

collector/src/main/java/com/bitmovin/analytics/data/DeviceInformationProvider.kt
```diff
@@ -4,20 +4,32 @@ import android.app.UiModeManager
...
 
 open class DeviceInformationProvider(val context: Context, val userAgent: String) {
     var isTV: Boolean = isTVDevice()
 
     fun getDeviceInformation(): DeviceInformation {
+
+        val windowManager: WindowManager = context.getSystemService(Context.WINDOW_SERVICE) as WindowManager
+        val displayMetrics = DisplayMetrics()
+
+        windowManager.defaultDisplay.getMetrics(displayMetrics)
```
rolandgriesser, line 20: Can WindowManager be `null`?

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Development
 
+### Changed
+- Way of getting available screen width and height
```
rolandgriesser, line 11: Should we mention that we changed that to include screen density?

collector/src/main/java/com/bitmovin/analytics/data/DeviceInformationProvider.kt
KulenovicInes: I believe that `WindowManager` is non nullable type. As long as `context` is not null, everything should be fine. `getSystemService()` method should return null only in cases when some random string is passed as name parameter, and service with that name doesn't exist.

We already use `context.getSystemService()` in `isTVDevice()` method, so I think it should be safe to use it here.

CHANGELOG.md
KulenovicInes: Changed in https://github.com/bitmovin/bitmovin-analytics-collector-android/pull/118/commits/ccd9a3fcaab2a09c15624b04ec51bd0cab7e062d. Does this maybe has more sense?

collector/src/main/java/com/bitmovin/analytics/data/DeviceInformationProvider.kt
Tigraine: probably should get a guard or something here

KulenovicInes: Added null check in https://github.com/bitmovin/bitmovin-analytics-collector-android/pull/118/commits/02c97d8446258f0af46b08139bf4fd3cebc34db6

PR #120, opened 2020-08-18
AN-1675 Added try-catch blocks on collectors

KulenovicInes:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1675
- Description: Added try-catch blocks around exoplayer and bitmovin listener methods. Didn't put them around all possible methods because in some cases methods have non nullable return type and in case of exceptions returning some default value like `false` or `0` would cause having wrong data in collector.

### Checklist

- [X] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - Calculations for available screen width and height to now include screen density and be more independent of physical pixel
-- Make collectors more resistant to exceptions
+- Make collectors more resilient to exceptions in event listeners
 
 ## v1.18.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - Calculations for available screen width and height to now include screen density and be more independent of physical pixel
+- Make collectors more resistant to exceptions
 
 ## v1.18.0
 
--- a/collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
+++ b/collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
@@ -28,7 +28,7 @@ class BitmovinSdkAdAdapter(val bitmovinPlayer: BitmovinPlayer, val adAnalytics: 
     private val adMapper: AdMapper = AdMapper()
     private val adBreakMapper: AdBreakMapper = AdBreakMapper()
     private val adQuartileFactory: AdQuartileFactory = AdQuartileFactory()
-    private val TAG = "BitmovinADPlayerAdapter"
+    private val TAG = "BitmovinSdkAdAdapter"
 
     private val onAdStartedListener = OnAdStartedListener {
         try {
@@ -118,39 +118,31 @@ class BitmovinSdkAdAdapter(val bitmovinPlayer: BitmovinPlayer, val adAnalytics: 
     }
 
     init {
-        try {
-            bitmovinPlayer.addEventListener(onAdStartedListener)
-            bitmovinPlayer.addEventListener(onAdFinishedListener)
-            bitmovinPlayer.addEventListener(onAdBreakStartedListener)
-            bitmovinPlayer.addEventListener(onAdBreakFinishedListener)
-            bitmovinPlayer.addEventListener(onAdClickedListener)
-            bitmovinPlayer.addEventListener(onAdErrorListener)
-            bitmovinPlayer.addEventListener(onAdSkippedListener)
-            bitmovinPlayer.addEventListener(onAdManifestLoadedListener)
-            bitmovinPlayer.addEventListener(onPlayListener)
-            bitmovinPlayer.addEventListener(onPausedListener)
-            bitmovinPlayer.addEventListener(onAdQuartileListener)
-        } catch (e: Exception) {
-            Log.d(TAG, "On Ad init", e)
-        }
+        bitmovinPlayer.addEventListener(onAdStartedListener)
+        bitmovinPlayer.addEventListener(onAdFinishedListener)
+        bitmovinPlayer.addEventListener(onAdBreakStartedListener)
+        bitmovinPlayer.addEventListener(onAdBreakFinishedListener)
+        bitmovinPlayer.addEventListener(onAdClickedListener)
+        bitmovinPlayer.addEventListener(onAdErrorListener)
+        bitmovinPlayer.addEventListener(onAdSkippedListener)
+        bitmovinPlayer.addEventListener(onAdManifestLoadedListener)
+        bitmovinPlayer.addEventListener(onPlayListener)
+        bitmovinPlayer.addEventListener(onPausedListener)
+        bitmovinPlayer.addEventListener(onAdQuartileListener)
     }
 
     override fun release() {
-        try {
-            bitmovinPlayer.removeEventListener(onAdStartedListener)
-            bitmovinPlayer.removeEventListener(onAdFinishedListener)
-            bitmovinPlayer.removeEventListener(onAdBreakStartedListener)
-            bitmovinPlayer.removeEventListener(onAdBreakFinishedListener)
-            bitmovinPlayer.removeEventListener(onAdClickedListener)
-            bitmovinPlayer.removeEventListener(onAdErrorListener)
-            bitmovinPlayer.removeEventListener(onAdSkippedListener)
-            bitmovinPlayer.removeEventListener(onAdManifestLoadedListener)
-            bitmovinPlayer.removeEventListener(onPlayListener)
-            bitmovinPlayer.removeEventListener(onPausedListener)
-            bitmovinPlayer.removeEventListener(onAdQuartileListener)
-        } catch (e: Exception) {
-            Log.d(TAG, "On Ad release", e)
-        }
+        bitmovinPlayer.removeEventListener(onAdStartedListener)
+        bitmovinPlayer.removeEventListener(onAdFinishedListener)
+        bitmovinPlayer.removeEventListener(onAdBreakStartedListener)
+        bitmovinPlayer.removeEventListener(onAdBreakFinishedListener)
+        bitmovinPlayer.removeEventListener(onAdClickedListener)
+        bitmovinPlayer.removeEventListener(onAdErrorListener)
+        bitmovinPlayer.removeEventListener(onAdSkippedListener)
+        bitmovinPlayer.removeEventListener(onAdManifestLoadedListener)
+        bitmovinPlayer.removeEventListener(onPlayListener)
+        bitmovinPlayer.removeEventListener(onPausedListener)
+        bitmovinPlayer.removeEventListener(onAdQuartileListener)
     }
 
     override val isLinearAdActive: Boolean
@@ -159,12 +151,5 @@ class BitmovinSdkAdAdapter(val bitmovinPlayer: BitmovinPlayer, val adAnalytics: 
         // TODO get actual module from player
         get() = AdModuleInformation("DefaultAdvertisingService", BitmovinUtil.getPlayerVersion())
     override val isAutoplayEnabled: Boolean?
-        get() {
-            return try {
-                bitmovinPlayer.config.playbackConfiguration?.isAutoplayEnabled
-            } catch (e: Exception) {
-                Log.d(TAG, "isAutoplayEnabled", e)
-                null
-            }
-        }
+        get() = bitmovinPlayer.config.playbackConfiguration?.isAutoplayEnabled
 }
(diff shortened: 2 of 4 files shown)


collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
```diff
@@ -76,35 +110,47 @@ class BitmovinSdkAdAdapter(val bitmovinPlayer: BitmovinPlayer, val adAnalytics:
...
+            bitmovinPlayer.addEventListener(onAdErrorListener)
+            bitmovinPlayer.addEventListener(onAdSkippedListener)
+            bitmovinPlayer.addEventListener(onAdManifestLoadedListener)
+            bitmovinPlayer.addEventListener(onPlayListener)
+            bitmovinPlayer.addEventListener(onPausedListener)
+            bitmovinPlayer.addEventListener(onAdQuartileListener)
+        } catch (e: Exception) {
+            Log.d(TAG, "On Ad init", e)
+        }
     }
```
sabe25, line 136: Do you think it is necessary here?
What could possibly go wrong?
same for release

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
```diff
@@ -113,5 +159,12 @@ class BitmovinSdkAdAdapter(val bitmovinPlayer: BitmovinPlayer, val adAnalytics:
...
     override val isAutoplayEnabled: Boolean?
-        get() = bitmovinPlayer.config.playbackConfiguration?.isAutoplayEnabled
+        get() {
+            return try {
+                bitmovinPlayer.config.playbackConfiguration?.isAutoplayEnabled
+            } catch (e: Exception) {
+                Log.d(TAG, "isAutoplayEnabled", e)
+                null
+            }
+        }
```
sabe25, line 169: Also i don't think we need a try-catch block in here

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
```diff
@@ -94,163 +94,182 @@ public void init() {
...
-        // DRM Information
-        if (drmInformation != null) {
-            data.setDrmType(drmInformation.getType());
+            return data;
+        } catch (Exception e) {
+            Log.d(TAG, "On Creating Event Data", e);
         }
-
-        return data;
+        return null;
```
sabe25, line 258: For me this move looks dangerous...
In `BitmovinAnalytics` we have no null check and therefore would throw another exception
Did we tracked any error during the creation of event Data?

KulenovicInes: This method gets a lot of data from player, so for me there is a great risk that something will possibly go wrong here. ATM if something goes wrong here it will probably be caught by other try catch blocks, but still I wouldn't like to rely on that some caller method will handle exceptions happening here.
I agree that returning null is not the best option, but maybe it's less harm then completely crashing app 🤷‍♀️ 
I'm open to any ideas here (also to remove this try catch completely)

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
KulenovicInes: Well we don't have any control of what is happening in `addEventListener` method. It's highly unlikely that something will go wrong here, but still I would prefer to be safe :)

KulenovicInes:
> I don't think it is necessary to add try catch blocks everywhere
> The main issue was that we added some event listeners to the player which crashed everything
> So all event listeners should have such a try-catch block
> 
> In addition we could make the statemachine a little bit more error resiliant by catching exceptions during `onExit` and `onEnter`
> although i'm not sure if we should catch the whole event or the listener calls within those methods

As far as I can see, first event listener is triggered, then state maschine, then `BitmovinAnalytics` with `createEventData()` etc. So wrapping event listener methods with try catch blocks we will catch also potential exceptions happening in state machine or in `BitmovinAnalytics`.
I'm ok with adding some additional try catch in statemachine, but then do we want to add some additional logic in `catch` part, like resetting timers?
ATM the biggest concern I have is that statemachine will get stuck in some state forever, but also doesn't make much sense to me to do some state transitions in catch part.... 🤷‍♀️

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
```diff
@@ -94,163 +94,182 @@ public void init() {
...
-        double duration = bitmovinPlayer.getDuration();
-        if (duration != Double.POSITIVE_INFINITY) {
-            data.setVideoDuration((long) duration * Util.MILLISECONDS_IN_SECONDS);
-        }
-
-        //ad
-        if (bitmovinPlayer.isAd()) {
-            data.setAd(1);
-        }
+        try {
```
rolandgriesser, line 162: Please remove here as well: This is Java code - we never expect something to be non-nullable here (check the line with video-quality for example: we get the object, and then check if it's null or not, but never pretend that it's non-nullable).
We need to make sure to also do that in future in our Java classes, and at one point, when we change to Kotlin, those things should be breaking changes in the player code. But I'd prefer this to crash then as well.
The reason why we fix this now can be seen in our Ads-code, as we use Kotlin there: (just an example, not sure if this was the exact reason:) Older player versions have the `AdBreakEvent` with a non-nullable `AdBreak`, which will translate to `val adBreak: AdBreak` in Kotlin. So in our auto-complete etc., we would directly access it, because obviously it cannot be null. Player-team introduced a change though, where they changed this to `val adBreak: AdBreak?`. So if you update the player, the code actually still works, because both would translate to `public AdBreak getAdBreak()` in Java - but in Java it can return `null`. This is where our code would crash if a newer player-version would return `null`.
So for now, let's just keep those try-catch blocks to the eventhandlers, not to everything IMO.
I'd like to have @Tigraine 's opinion as well though, before you change anything.

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
```diff
@@ -94,163 +94,182 @@ public void init() {
...
-        return data;
+        return null;
     }
 
     @Override
     public void release() {
-        playerIsReady = false;
-        if (bitmovinPlayer != null) {
-            removePlayerListener();
+        try {
```
rolandgriesser, line 263: can be removed

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
```diff
@@ -274,18 +293,27 @@ public void clearValues() {
       so in case the player starts a video due to autoplay=true we need to transition into startup state manually
      */
     private void checkAutoplayStartup() {
-        PlaybackConfiguration playbackConfiguration = bitmovinPlayer.getConfig().getPlaybackConfiguration();
-        SourceConfiguration source = bitmovinPlayer.getConfig().getSourceConfiguration();
-        if (playbackConfiguration != null && source != null && source.getFirstSourceItem() != null && playbackConfiguration.isAutoplayEnabled()) {
-            startup();
+        try {
```
rolandgriesser, line 296: can be removed, would maybe include a `bitmovinPlayer.getConfig() != null` check though (shouldn't be nullable, but in case that changes)

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
```diff
@@ -274,18 +293,27 @@ public void clearValues() {
...
+        } catch (Exception e) {
+            Log.d(TAG, "On Check Autoplay Startup", e);
         }
     }
 
     private void startup() {
-        stateMachine.transitionState(PlayerState.STARTUP, getPosition());
-        if (!bitmovinPlayer.isAd()) {
-            isVideoAttemptedPlay = true;
+        try {
```
rolandgriesser, line 308: Can be removed

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
```diff
@@ -303,30 +331,42 @@ public void onSourceLoaded(SourceLoadedEvent sourceLoadedEvent) {
...
     };
 
     private OnDestroyListener onDestroyedListener = new OnDestroyListener() {
         @Override
         public void onDestroy(DestroyEvent destroyEvent) {
-            Log.d(TAG, "On Destroy");
-            if (!stateMachine.isStartupFinished() && isVideoAttemptedPlay) {
-                stateMachine.setVideoStartFailedReason(VideoStartFailedReason.PAGE_CLOSED);
-                stateMachine.transitionState(PlayerState.EXITBEFOREVIDEOSTART, getPosition());
+            try{
```
rolandgriesser, line 346: Code style

collector-exoplayer/src/main/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapter.java
```diff
@@ -124,34 +124,43 @@ public void init() {
 
     @Override
     public EventData createEventData() {
-        EventData data = factory.build(stateMachine.getImpressionId());
+        try {
```
rolandgriesser, line 127: Can be removed

collector-exoplayer/src/main/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapter.java
```diff
@@ -124,34 +124,43 @@ public void init() {
...
     public void release() {
-        playerIsReady = false;
-        manifestUrl = null;
-        if (this.exoplayer != null) {
-            this.exoplayer.removeListener(this);
-        }
-        if (this.exoplayer instanceof SimpleExoPlayer) {
-            SimpleExoPlayer simpleExoPlayer = (SimpleExoPlayer) this.exoplayer;
-            simpleExoPlayer.removeAnalyticsListener(this);
+        try {
```
rolandgriesser, line 149: can be removed

collector-exoplayer/src/main/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapter.java
```diff
@@ -203,58 +212,66 @@ public void onLoadingChanged(boolean isLoading) {
 
     @Override
     public void onPlayerStateChanged(boolean playWhenReady, int playbackState) {
-        long videoTime = getPosition();
-        Log.d(TAG, String.format("onPlayerStateChanged: %b, %s", playWhenReady, ExoUtil.exoStateToString(playbackState)));
+        try {
```
rolandgriesser, line 215: I'm not sure if we need that at all for Exoplayer. Again, @Tigraine , any opinion?

rolandgriesser (changes requested):

collector-exoplayer/src/main/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapter.java
Tigraine: I'd add it anyways .. We really want to make sure that the collector never crashes the App 🤔 . So when in doubt: add more try/catch 😆

sabe25 (approved):

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdAdapter.kt
```diff
@@ -27,44 +28,77 @@ class BitmovinSdkAdAdapter(val bitmovinPlayer: BitmovinPlayer, val adAnalytics:
...
+    private val TAG = "BitmovinSdkAdAdapter"
 
     private val onAdStartedListener = OnAdStartedListener {
-        val ad = it.ad ?: return@OnAdStartedListener
-        adAnalytics.onAdStarted(adMapper.fromPlayerAd(ad))
+        try {
+            val ad = it.ad ?: return@OnAdStartedListener
+            adAnalytics.onAdStarted(adMapper.fromPlayerAd(ad))
+        } catch (e: Exception) {
+            Log.d(TAG, "On Ad Started", e)
```
Tigraine, line 38: Can't we just directly log the exception? Or some more descriptive text - because the Stacktrace should contain the location anyways so no need to repeat ourselves

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - Calculations for available screen width and height to now include screen density and be more independent of physical pixel
+- Make collectors more resistant to exceptions
```
Tigraine, line 12:
```suggestion
- Make collectors more resilient to exceptions in event listeners
```

Tigraine (approved):
LGTM aside from one comment that I am not sure if I like the Log texts

rolandgriesser (approved):

PR #127, opened 2020-11-02
Feature/deploy script improvements

rolandgriesser:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1880
- Description: Deploy script didn't override `developLocal` properly. Changed setting the `developLocal` in the deploy script to a system property.

I'm pretty confident this works now, but as this can only be really tested with a real release, let's take a close look at the next time we release something.
Also, when releasing, we should try to debug both exoplayer and bitmovin player at least once, manually setting `developLocal` to `false` and checking if the outgoing `version` includes the `-local` or not (if it doesn't, it should correctly have pulled the version from artifactory).

### Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,8 +40,9 @@ allprojects {
 
 ext {
     developLocal =  System.getProperty("developLocal", "true").toBoolean()
+    versionNameSuffix = developLocal ? '-local' : ''
     packageGroupId = 'com.bitmovin.analytics'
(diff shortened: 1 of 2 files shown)


sabe25 (approved):
Can we also adapt the Readme so everybody knows for what to look for when doing releases?

Tigraine (approved):

PR #129, opened 2020-11-05
AN-1812 Segment Tracking

rolandgriesser:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1812
- Description: Added segment tracking and error details collection

### Checklist

- [x] Updated CHANGELOG.md


collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
```diff
@@ -78,15 +89,27 @@
...
+        this.analytics = analytics;
         this.config = config;
         this.stateMachine = stateMachine;
         this.bitmovinPlayer = bitmovinPlayer;
         this.factory = factory;
+        this.context = context;
     }
 
     public void init() {
+        this.segmentTrackingAdapter = new BitmovinSegmentTrackingAdapter(this.bitmovinPlayer);
```
Tigraine, line 107: i'd move all of this out of here into a factory. it will get more and more over time so starting to seperate object creation from usage would be a good idea anyways.
Might make sense to look at DI Containers that manage object lifecycle like Guice or this other thing that the player ppl are starting to use on Android

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
```diff
@@ -247,10 +270,17 @@ public EventData createEventData() {
     @Override
     public void release() {
         playerIsReady = false;
+        if(this.segmentTrackingAdapter != null) {
```
Tigraine, line 273: Long term we should build a analytics event-bus so the feature can hook up to a `release` event and we don't have to manually release each feature in the main adapter

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
```diff
@@ -206,7 +218,7 @@ public void onSeekComplete(long duration) {
         data.setDuration(duration);
         data.setVideoTimeStart(playerStateMachine.getVideoTimeStart());
         data.setVideoTimeEnd(playerStateMachine.getVideoTimeEnd());
-        sendEventData(data);
+        sendEventData(data, playerStateMachine.getCurrentState(), PlayerEvent.SEEKED);
```
Tigraine, line 221: i'm in two minds about exposing the state and the event.. we are exposing a bunch of information to the features that might be useful but also might lead to unintended uses down the line if we are not careful.

collector/src/main/java/com/bitmovin/analytics/features/segmenttracking/SegmentTrackingFeature.kt
```diff
@@ -0,0 +1,50 @@
...
+
+import com.bitmovin.analytics.data.AdEventData
+import com.bitmovin.analytics.data.EventData
+import com.bitmovin.analytics.features.EventSource
+import com.bitmovin.analytics.features.Feature
+import com.bitmovin.analytics.stateMachines.PlayerEvent
+import com.bitmovin.analytics.stateMachines.PlayerState
+import java.util.*
+
+class SegmentTrackingFeature(private vararg val eventSources: EventSource<SegmentTrackingEventListener>) : Feature<SegmentTrackingFeatureConfig>(), SegmentTrackingEventListener {
```
Tigraine, line 11: we might want to drop the `Feature` part as I see us structuring most of the new code around these "feature" buckets anyways.. so a SegmentTracker that then gets used by multiple other things again .. 🤔

collector/src/main/java/com/bitmovin/analytics/features/segmenttracking/SegmentTrackingFeatureConfig.kt
```diff
@@ -0,0 +1,7 @@
+package com.bitmovin.analytics.features.segmenttracking
+
+import com.bitmovin.analytics.features.FeatureConfig
+
+class SegmentTrackingFeatureConfig: FeatureConfig() {
+    val maxSegments: Int = 20
```
Tigraine, line 6: let's start with 10 for the beginning
```suggestion
    val maxSegments: Int = 10
```

Tigraine (commented):
looks good - I think we need an Event Bus 🙈

Svarozic (changes requested):
CodeShip build fails

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
rolandgriesser: Do we really want to pull in external dependencies?
How would that look like with a factory? Just something simple like a `BitmovinFeaturesFactory.create(player): List<Features>` where we create all that stuff? So only moving it out from here?

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
rolandgriesser: Ok, will remove that part completely for now.
But the only alternative I can think of for new would be something like hooking the features up to actual events like: `onBeforeExitErrorState`, `onBeforeExitBufferingState` etc. so we can set things on specific events. Which looks quite ugly, but you get the idea --> would look better once we restructure those `onExit` things.

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
rolandgriesser: Could you check the last two commits for that? It's a bit ugly right now as the eventbus is really simple - but it should do what we want and for this beta release should be fine I guess?
I'd create a separate issue to make a streamlined event bus here.

collector/src/main/java/com/bitmovin/analytics/features/errordetails/ErrorDetail.kt
```diff
@@ -0,0 +1,5 @@
+package com.bitmovin.analytics.features.errordetails
+
+import com.bitmovin.analytics.features.segmenttracking.Segment
+
+data class ErrorDetail(val timestamp: Long, val code: Int?, val message: String?, val stacktrace: Array<String>?, val segments: Collection<Segment>?)
```
Svarozic, line 5: why `Array<String>` ? why not listOf ? do we plan to change array items in runtime ?

collector/src/main/java/com/bitmovin/analytics/features/FeatureManager.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.bitmovin.analytics.data.AdEventData
+import com.bitmovin.analytics.data.EventData
+
+class FeatureManager {
+    companion object {
+        val TAG = FeatureManager::class.java.name
+    }
+    private val features: MutableList<Feature<*>> = mutableListOf()
+
+    fun registerFeature(feature: Feature<*>) {
```
Svarozic, line 13: this method is not used, and as I understand it is entry point to the whole feature tracking ? how is then code from this PR used ?

collector/src/main/java/com/bitmovin/analytics/PlayerAdapterBase.kt
```diff
@@ -0,0 +1,15 @@
+package com.bitmovin.analytics
+
+import com.bitmovin.analytics.features.Feature
+
+open class PlayerAdapterBase {
```
Svarozic, line 5: why is it not `abstract` ? do we expect instances of it ?

collector/src/main/java/com/bitmovin/analytics/features/FeatureFactory.kt
```diff
@@ -0,0 +1,5 @@
+package com.bitmovin.analytics.features
+
+interface FeatureFactory {
+    fun createFeatures(): Collection<Feature<*>>
+}
```
Svarozic, line 5: I dont see a big value of this interface and its implementations, as adapters could have same method `createFeatures` (defined by base class) that would produce list of features. so in `PlayerAdapterBase` I would implement `registerFeatures` as abstract method, that returns features and has to be implemented

collector/src/main/java/com/bitmovin/analytics/features/FeatureManager.kt
```diff
@@ -0,0 +1,32 @@
...
+
+import android.util.Log
+import com.bitmovin.analytics.data.AdEventData
+import com.bitmovin.analytics.data.EventData
+
+class FeatureManager {
+    companion object {
+        val TAG = FeatureManager::class.java.name
+    }
+    private val features: MutableList<Feature<*>> = mutableListOf()
```
Svarozic, line 11: I see list of features here and in `PlayerAdapterBase` , what is a difference ?

Svarozic (commented):
I gave it just brief look, as I had a problem to understand, how all classes are connected to each other

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
```diff
@@ -101,6 +108,9 @@ protected void attachAd(AdAdapter adapter) {
     public void detachPlayer() {
         detachAd();
```
Tigraine, line 109: food for thought for the future - can we move the ad to the `featureManager` infrastructure?

PR #130, opened 2020-11-19
AN-1919 Override mpUrl and m3u8Url with AnalyticsConfiguration

peterpf:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1919
- Description:
    - Refactoring of `EventData` decoration: use a `EvenDataDecorationPipeline` and register pre/post decorators to set default values or override values of the data that is received from the adapter.
    - Added post-`EventDataDecorator` that overrides `mpdUrl` and `m3u8Url` with values from the `AnalyticsConfiguration`
    - Updated exoplayer/bitmovinPlayer examples to use detach/attach procedure before changing source 

### Checklist

- [x] Updated CHANGELOG.md


KulenovicInes (approved):
Looks good to me 👍

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
```diff
@@ -206,6 +206,13 @@ public EventData createEventData() {
                     break;
             }
         }
+        // Override mpdUrl and m3u8 url if they are set in the AnalyticsConfig
```
rolandgriesser, line 209: @Tigraine what should have precedence here? 
I think the config values should be the fallbacks, but if we can actually get that information from the player, we should use that?

@peterpf  depending on Daniel's answer, we could also always set this in the `EventData` constructor already and this can be independent of the SDK (so we don't have to do it in both). Or even if that's not the case, we could try to set those in one place, as it's duplicate code otherwise

Tigraine: @rolandgriesser  Actually the order of precedence here is config > player.. Because the stuff we get from the player is actually wrong

rolandgriesser: ah, I thought that's only for the cases where we don't get the information. Good to know - still should be done in the base collector IMO, as it's the same functionality, no matter the SDK

Tigraine: yes - agree with the duplicate code.

But this is for RTL who are struggling with the streamRoot SDK setting 127.0.0.1 as MPD-URL so they want to be able to configure the "real" URL regardless of proxy.

peterpf: I was thinking about putting the code in the `sendEventData` in the analytics core, but was not sure if this is the right place. @rolandgriesser what do you think where the best place is to put this common code?

rolandgriesser: Might need to refactor a bit here, as the `createEventData` from the adapter is also used in the ads - so putting this in `sendEventData` would not apply to ads. A way would be to have a `PlayerAdapter` base class that has a `createEventData` and an abstract `decorateEventData`, so we would create the eventData in the base class and the adapters would only decorate it then. 
@Tigraine thoughts?

Tigraine: please do a pair programming session on this :)

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalyticsConfig.java
```diff
@@ -286,10 +298,27 @@ public void setTitle(String title) {
     }
 
     /**
-     * A/B Test Experiment Name
+     * MPD URL of the manifest. Usually auto-detected, but can be overwritten.
```
Tigraine, line 301: i'd change the comment to say "If not set explicitly the collector will retrieve available information from the player"

collector/src/main/java/com/bitmovin/analytics/data/EventData.kt
```diff
@@ -7,14 +7,14 @@ import com.bitmovin.analytics.utils.Util
...
     val impressionId: String,
-    deviceInfo: DeviceInformation,
     val userId: String
 ) {
-    val userAgent = deviceInfo.userAgent
-    val deviceInformation = DeviceInformationDto(deviceInfo.manufacturer, deviceInfo.model, deviceInfo.isTV)
-    val language: String = deviceInfo.locale //
+    var userAgent: String? = null
+    var deviceInformation: DeviceInformationDto? = null
+    var language: String? = null
```
Tigraine, line 14: this was non-nullable before?

collector/src/main/java/com/bitmovin/analytics/data/EventData.kt
```diff
@@ -69,7 +70,7 @@ class EventData(
     var progUrl: String? = null
     var isMuted = false
     var sequenceNumber: Int = 0
-    val platform: String = if (deviceInfo.isTV) "androidTV" else "android"
+    var platform: String? = null
```
Tigraine, line 73: Let's please not make this nullable! This just leads to someone forgetting some hook somewhere and we have null being sent to the server :/

collector/src/main/java/com/bitmovin/analytics/data/EventData.kt
peterpf: Those parameters have been set in the constructor of the `EventData` before. This was moved to one of the `EventDataDecoratorPipeline`s we introduced in this PR. As a consequence this was changed to be `nullable`.

collector/src/main/java/com/bitmovin/analytics/data/EventData.kt
peterpf: Considering https://github.com/bitmovin/bitmovin-analytics-collector-android/pull/130#discussion_r533278132, do you see a quick/nice way of working around this?

collector/src/main/java/com/bitmovin/analytics/data/EventData.kt
Tigraine: :/ this looks like something that's very prone to mistakes / errors or unintended consequences long term :( 
@rolandgriesser what do you think?

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalyticsConfig.java
peterpf: Changed in 8e781300a0c05a4b7dd3ac471563976ee9db58e8

collector/src/main/java/com/bitmovin/analytics/data/EventData.kt
rolandgriesser: Those values were only non-nullable as we retrieved them in the constructor before. I would rather go with one system to *decorate* samples instead of having multiple ways. 
Something we could do: have something a `DeviceInformationDecorator` interface and then after calling the `registerDecorators`, check if this decorator has been provided by the player adapter and otherwise throw an exception. And then pass this decorator in the constructor of `EventData` (and retrieve the properties). But technically we could do this with every available decorator and make them nullable, which would lead to huge constructors and `optional` stuff would be handled really strange.

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalyticsConfig.java
```diff
@@ -298,7 +298,8 @@ public void setTitle(String title) {
     }
 
     /**
-     * MPD URL of the manifest. Usually auto-detected, but can be overwritten.
+     * MPD URL of the manifest.
```
Tigraine, line 301: nit pick: the method is setting the url - your comments reads like a getter ;)

collector/src/main/java/com/bitmovin/analytics/data/EventData.kt
peterpf: Undid changes to `EventData` in 4a51afec249f3ec25ff41013b01c4ffb321b4ff7 (pass deviceInformation to `EventData`) and renamed `decorators` to `manipulators`

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinPlayerCollector.java
```diff
@@ -32,8 +30,8 @@ public BitmovinPlayerCollector(
     }
 
     public void attachPlayer(BitmovinPlayer player) {
-        EventDataFactory factory = new EventDataFactory(this.bitmovinAnalyticsConfig, context, new DeviceInformationProvider(context, getUserAgent(context)), new UserIdProvider(context));
-        BitmovinSdkAdapter adapter = new BitmovinSdkAdapter(player, this.bitmovinAnalyticsConfig, factory,
+        DeviceInformationProvider deviceInformationProvider = new DeviceInformationProvider(context, getUserAgent(context));
```
rolandgriesser, line 33: @Tigraine Short explanation: as we want the `EventData` creation to happen in `BitmovinAnalytics`, and not in the player adapter, we will remove the factory here for now (to keep things simple), and takle that in a separate PR.

collector-bitmovin-player/src/main/java/com/bitmovin/analytics/bitmovin/player/BitmovinSdkAdapter.java
```diff
@@ -149,10 +152,7 @@ private void removePlayerListener() {
     }
 
     @Override
-    public EventData createEventData() {
-        EventData data = factory.build(stateMachine.getImpressionId());
-
-        data.setAnalyticsVersion(BuildConfig.VERSION_NAME);
```
rolandgriesser, line 155: Please test this in detail: Before it was set in both the `EventData` class and here. Maybe this has something to do with gradle package imports. (I guess it shouldn't be a problem as we can't have Core 1.2.0 with BitmovinCollector 1.3.0 etc., but please check the outgoing payload if everything works correctly.

peterpf: When changing the value in `build.gradle (Project: collector-android)` it will set this as the `analyticsVersion` in the payload. Tested locally with bitmovin and exoplayer example.

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalyticsConfig.java
peterpf: Fixed in c6b4cab34ab6a0906c62e470935495470f95aa19

PR #131, opened 2020-12-04
Prevent serialisation of errorDetails as `null` string

peterpf:
### Work
- Fixes: 
- Description: See slack discussion https://bitmovin.slack.com/archives/CE8HQ8SBY/p1607077655009200. Error Details have been serialised as `null` string.

### Checklist

- [ ] Updated CHANGELOG.md


collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
```diff
@@ -221,7 +221,9 @@ public void onError(ErrorCode errorCode) {
 
         data.setErrorCode(errorCode.getErrorCode());
         data.setErrorMessage(errorCode.getDescription());
-        data.setErrorData(serialize(errorCode.getErrorData()));
```
rolandgriesser, line 224: Would it make sense to change this in the `serialize` method? If an object is `null`, we never want it to be serialized to the string I think? Not sure in which places we use that method, but I can't think of a reason where we want to send `"null"` to the server, and if it's really intended, I think it would make more sense to implement the opposite case: `val serialized = (obj == null) ? "null": serialize(obj)`.

sabe25: Well i would expect this method to return "null" for null because it should just serialize it
what about adding a new method `serializeOrNull(Any): String?` which is clear on the return type?

peterpf: So far I have only seen this instance where we explicitly serialise an object.
Regarding your suggestion:
What about `customDataX` fields? We usually don't set them to safe storage.

rolandgriesser: TBH I would not expect a method to serialize `null` to a `"null"` string. I don't hink if you are the consumer of the serialized string, you'd want to start parsing a `"null"` string, instead you'd expect the object to be set to `null` already.
@Tigraine thoughts?

Tigraine: I also don't think sending a string `"null"` to the server makes any sense TBH.
If the `errorDetails` are null then they should be null. This is better then addressed at the presentation layer.

rolandgriesser: But where would you but the check? I'd expect the serialize method to not do anything for null (only return null), instead of having another method that wraps it

Tigraine (approved):
LGTM

collector/src/main/java/com/bitmovin/analytics/utils/DataSerializer.kt
```diff
@@ -0,0 +1,27 @@
...
+        if (data == null) {
+            return null
+        }
+
+        val gson = Gson()
+        return gson.toJson(data)
+    }
+
+    @JvmStatic
+    fun <T> deserialize(input: String?, classOfT: Class<T>?): T? {
```
rolandgriesser, line 18:
```suggestion
fun <T> deserialize(input: String?, classOfT: Class<T>): T? {
        input ?: return null
        return try {
            Gson().fromJson(input, classOfT)
        } catch (e: JsonSyntaxException) {
            null
        }
    }
```

collector/src/main/java/com/bitmovin/analytics/utils/DataSerializer.kt
```diff
@@ -0,0 +1,27 @@
+package com.bitmovin.analytics.utils
+
+import com.google.gson.Gson
+import com.google.gson.JsonSyntaxException
+
+object DataSerializer {
+    @JvmStatic
+    fun <T> serialize(data: T?): String? {
```
rolandgriesser, line 8:
```suggestion
fun <T> serialize(data: T?): String? {
        data ?: return null
        return Gson().toJson(data)
    }
```

rolandgriesser (changes requested):

collector/src/main/java/com/bitmovin/analytics/utils/DataSerializer.kt
peterpf: The `@JvmStatic` annotations are necessary for the java classes that access those methods

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
peterpf: Resolved by adding the null check into the serialisation method

rolandgriesser (approved):

PR #132, opened 2020-12-14
AN-1791 cache samples

KulenovicInes:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1791
- Description: 
   - Added logic for retry in duration of 5minutes
   - Added config option
   - Updated http client to remove its own retry policy because that only gets samples multiplied on backend

### Checklist
- [ ] Updated CHANGELOG.md


collector/src/main/java/com/bitmovin/analytics/BitmovinAnalyticsConfig.java
```diff
@@ -30,6 +30,7 @@
     private Boolean ads = true;
     private Context context;
     private Boolean isLive;
+    private Boolean resendDataOnHttpTimeout = false;
```
rolandgriesser, line 33: Not sure about the name of this config, as it's really specific und would be hard to change later (without breaking changes). Do we want to have that more generic, what are our plans for this feature in the future? (in regards to supporting longer offline sessions?) @Tigraine

collector/src/main/java/com/bitmovin/analytics/data/Backend.kt
```diff
@@ -0,0 +1,8 @@
+package com.bitmovin.analytics.data
+
+import okhttp3.Callback
+
+interface Backend {
+    fun send(eventData: EventData, callback: Callback?)
```
rolandgriesser, line 6: is there another way this could be done? the `RetryBackend` has this parameter as well but will just ignore it -  I would expect the callback to be called though.
So maybe the `inner` backend for the `RetryBackend` should actually be a `HttpBackend`, and have those 2 methods, or we should implement it properly in the `RetryBackend` as well.

analyticsexample/src/main/java/com/bitmovin/exoplayeranalyticsexample/MainActivity.java
```diff
@@ -94,6 +94,7 @@ private void createPlayer() {
             bitmovinAnalyticsConfig.setPath("/vod/new/");
             bitmovinAnalyticsConfig.setHeartbeatInterval(59700);
             bitmovinAnalyticsConfig.setIsLive(false);
+            bitmovinAnalyticsConfig.setResendDataOnHttpTimeout(true);
```
Tigraine, line 97: I think the placement of this setting should probably change to be moved to the `CollectorConfig`

collector/build.gradle
```diff
@@ -50,6 +58,12 @@ dependencies {
     androidTestImplementation "org.assertj:assertj-core:${rootProject.ext.assertjCoreVersion}"
     testImplementation "org.mockito:mockito-core:${rootProject.ext.mockitoCoreVersion}"
     testImplementation "com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0"
+
+
+    testImplementation "org.powermock:powermock-core:2.0.9"
```
Tigraine, line 63: why do we need powermock? Can't we do this with mockito alone?

collector/src/main/java/com/bitmovin/analytics/data/AdEventData.kt
```diff
@@ -110,7 +110,8 @@ data class AdEventData(
     var autoplay: Boolean? = null,
     var platform: String? = null,
     var audioCodec: String? = null,
-    var videoCodec: String? = null
+    var videoCodec: String? = null,
+    var retry: Int = 0
```
Tigraine, line 114: `retryCount` would be a better name

collector/src/main/java/com/bitmovin/analytics/data/Backend.kt
```diff
@@ -0,0 +1,8 @@
+package com.bitmovin.analytics.data
+
+import okhttp3.Callback
```
Tigraine, line 3: Would it be easy to decouple this interface from `okhttp3`?

collector/src/main/java/com/bitmovin/analytics/data/EventData.kt
```diff
@@ -83,4 +83,5 @@ class EventData(
     var videoStartFailed: Boolean = false
     var videoStartFailedReason: String? = null
     var downloadSpeedInfo: DownloadSpeedInfo? = null
+    var retry: Int = 0
```
Tigraine, line 86: `retryCount` would be a better name.

collector/src/main/java/com/bitmovin/analytics/data/HttpBackend.java
```diff
@@ -4,42 +4,43 @@
...
 import com.bitmovin.analytics.utils.HttpClient;
+import okhttp3.Callback;
 
 public class HttpBackend implements Backend {
     private static final String TAG = "BitmovinBackend";
     private HttpClient httpClient;
     private String analyticsBackendUrl;
     private String adsAnalyticsBackendUrl;
 
-    public HttpBackend(CollectorConfig config, Context context) {
```
Tigraine, line 17: I think you can stay with the `CollectorConfig` here as that's the configuration that configures how the collector works

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,162 @@
...
+
+    override fun sendAd(eventData: AdEventData, callback: Callback?) {
+        scheduleSample(RetrySample(null, eventData, 0, Date(), 0))
+    }
+
+    private fun scheduleSample(retrySample: RetrySample) {
+
+        Log.d(TAG, "sending sample retry ${retrySample.retry}")
+
+        if (retrySample.eventData != null) {
```
rolandgriesser, line 44: @Tigraine not sure about that part. Maybe we should separate between `ads` and normal samples and not have it in one list?
@KulenovicInes maybe make the `RetrySample` generic with a `data` property, and then handle it accordingly. This would be easier to extend in the future, now it's really specific to `EventData` and `AdEventData`.

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,162 @@
...
+                override fun onResponse(call: Call, response: Response) {
+                }
+            })
+        } else if (retrySample.adEventData != null) {
+            retrySample.adEventData.retry = retrySample.retry
+            this.next.sendAd(retrySample.adEventData, object : Callback {
+                override fun onFailure(call: Call, e: IOException) {
+
+                    if (e is SocketTimeoutException) {
+                        addSample(retrySample)
```
rolandgriesser, line 64: why is the part for Ads so different to the normal one? (e.g. `onResponse` having the `processQueuedSamples`, or having a `call.cancel()` in the other one?)

collector/src/main/java/com/bitmovin/analytics/utils/HttpClient.java
```diff
@@ -19,15 +22,26 @@
...
         this.context = context;
+        this.analyticsConfig = analyticsConfig;
     }
 
 
     public void post(String url, String postBody, final Callback callback) {
         if (client == null) {
-            client = new OkHttpClient();
+            if (analyticsConfig.isResendDataOnHttpTimeout()){
+                client = new OkHttpClient.Builder()
```
Tigraine, line 36: Hmm.. aren't these sane settings regardless of the retry feature? Why do we need to pass the configuration in here - probably the httpClient should be used outside of this class and be injected here?

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetrySample.kt
```diff
@@ -0,0 +1,16 @@
+package com.bitmovin.analytics.retryBackend
+
+import com.bitmovin.analytics.data.AdEventData
+import com.bitmovin.analytics.data.EventData
+import java.util.Date
+
+class RetrySample(val eventData: EventData?, val adEventData: AdEventData?, var totalTime: Int, var scheduledTime: Date, var retry: Int) : Comparable<RetrySample> {
+    override fun compareTo(other: RetrySample): Int {
```
Tigraine, line 8: This is a bit misleading as `compareTo` sounds like it's actually comparing the two samples, whereas this only compares the scheduleTimes. This could lead to all kinds of issues down the road if someone doesn't bother to read the implementation of `compareTo`

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,162 @@
...
+    private val lock = ReentrantLock()
+    private var retrySamplesSet = sortedSetOf<RetrySample>()
+    private var retryToken: Date? = null
+
+    override fun send(eventData: EventData, callback: Callback?) {
+        scheduleSample(RetrySample(eventData, null, 0, Date(), 0))
+    }
+
+    override fun sendAd(eventData: AdEventData, callback: Callback?) {
+        scheduleSample(RetrySample(null, eventData, 0, Date(), 0))
```
Tigraine, line 37: if you add a overridable func as a field in this class that returns `Date()` you can easily test `Date()`.
Something like `var now: () -> Date = fun() : Date { return Date() }`
This way you can't reliably set up `Date()` in tests.

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetrySample.kt
```diff
@@ -0,0 +1,16 @@
...
+        return when {
+            (this == null && other == null) -> 0
```
rolandgriesser, line 10: `this` can never be null here.
Also, as `scheduledTime` is not nullable, it would probably be enough to have 
`return this.scheduledTime.compareTo(other?.scheduledTime)`.

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,162 @@
...
+            Log.d(TAG, "addSample exception ${e.message}")
+        } finally {
+            lock.unlock()
+        }
+    }
+
+    private fun getSample(): RetrySample? {
+        try {
+            lock.lock()
+            val retrySample = retrySamplesSet.firstOrNull { it.scheduledTime.before(Date()) || it.scheduledTime == Date() }
```
rolandgriesser, line 146:
```suggestion
            val retrySample = retrySamplesSet.firstOrNull { it.scheduledTime.compareTo(Date()) <= 0 }
```
or `>=`, not sure now, should also do the trick and is easier to read.

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,162 @@
...
+                processQueuedSamples()
+            }
+        } catch (e: Exception) {
+            Log.d(TAG, "addSample exception ${e.message}")
+        } finally {
+            lock.unlock()
+        }
+    }
+
+    private fun getSample(): RetrySample? {
```
rolandgriesser, line 143: I'd give this a better name as we don't just try to get any sample. maybe in comparison to the `firstOrNull`, something like `nextSampleOrNull`

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,162 @@
...
+                    if (e is SocketTimeoutException || e is ConnectException || e is StreamResetException || e is UnknownHostException) {
+                        addSample(retrySample)
+                        call.cancel()
+                    }
+                }
+
+                override fun onResponse(call: Call, response: Response) {
+                }
+            })
+        } else if (retrySample.adEventData != null) {
```
Tigraine, line 58: early return would reduce nesting

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,162 @@
...
+            retrySample.retry++
+            val backOffTime = minOf(2.toDouble().pow(retrySample.retry).toInt(), 64)
+            // more than 5min in queue
+            if (retrySample.totalTime + backOffTime < MAX_RETRY_TIME) {
+
+                retrySample.scheduledTime = Calendar.getInstance().run {
+                    add(Calendar.SECOND, backOffTime)
+                    time
+                }
+                retrySample.totalTime += backOffTime
```
rolandgriesser, line 123: you can just move that before the `if`, as we will discard the sample anyways if it's too old - so you don't have to do the calculation twice (I know it's just a simple `+`, but in case it changes in future)

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
Tigraine, line 37: not sure if I like having RetrySample be "generic" for two types. 
Can't this be changed to not have null checks? What if we ever end up with a third "type"?

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,162 @@
...
+            this.next.send(retrySample.eventData, object : Callback {
+
+                override fun onFailure(call: Call, e: IOException) {
+                    if (e is SocketTimeoutException || e is ConnectException || e is StreamResetException || e is UnknownHostException) {
+                        addSample(retrySample)
+                        call.cancel()
+                    }
+                }
+
+                override fun onResponse(call: Call, response: Response) {
```
Tigraine, line 55: Why is `sendAd` `onResponse` calling `processQueuedSamples()` and the "normal" path isn't?

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,162 @@
...
+import okhttp3.Callback
+import okhttp3.Response
+import okhttp3.internal.http2.StreamResetException
+
+class RetryBackend(private val next: Backend, val handler: Handler) : Backend {
+
+    private val TAG = "RetryBackend"
+    private val lock = ReentrantLock()
+    private var retrySamplesSet = sortedSetOf<RetrySample>()
+    private var retryToken: Date? = null
```
Tigraine, line 30: Why is this called `token` but is a date?

collector/src/test/java/com/bitmovin/analytics/retryBackend/RetryBackendTest.kt
```diff
@@ -0,0 +1,188 @@
...
+
+        val firstSample = setupEventData(1)
+        val secondSample = setupEventData(2)
+
+        retryBacked.addSample(RetrySample(firstSample, null, 0, thirdDate, 0))
+        retryBacked.addSample(RetrySample(secondSample, null, 0, firstDate, 0))
+
+        // wait to be sure that processing of samples is finished
+        TimeUnit.SECONDS.sleep(2)
+        var sample = Whitebox.invokeMethod<RetrySample>(retryBacked, "getSample")
```
Tigraine, line 81: Don't you think this might be an indication that the class can be broken down further to avoid having to mock internals here?
What if the queue were to be moved outside of the `RetryBackend`? Wouldn't that simplify also testing and the synchronization logic?

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,162 @@
...
+            // more than 5min in queue
+            if (retrySample.totalTime + backOffTime < MAX_RETRY_TIME) {
+
+                retrySample.scheduledTime = Calendar.getInstance().run {
+                    add(Calendar.SECOND, backOffTime)
+                    time
+                }
+                retrySample.totalTime += backOffTime
+
+                if (retrySamplesSet.size > MAX_RETRY_SAMPLES) {
```
rolandgriesser, line 125: can you move that into it's on method? something like `truncateSamples`, you can even have a loop `while(size> max)` there, in case it's possible to add multiple samples at one point

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+    @Synchronized
+    fun processQueuedSamples() {
+        try {
+            if (retrySamplesSet.size > 0) {
+                val nextScheduledTime = retrySamplesSet.first().scheduledTime
+
+                if (retryToken != nextScheduledTime) {
```
rolandgriesser, line 81: what happens if two samples are scheduled at the same time?

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,162 @@
...
+                    if (retryToken != null) {
+                        handler.removeCallbacks(processSampleRunnable, retryToken)
+                    }
+                    retryToken = nextScheduledTime
+                    val delay = maxOf(nextScheduledTime.time - Date().time, 0) // to prevent negative delay
+                    val t = handler.postAtTime(processSampleRunnable, retryToken, SystemClock.uptimeMillis() + delay)
```
rolandgriesser, line 87: you don't use the `t`, can be removed.
```suggestion
handler.postAtTime(processSampleRunnable, retryToken, SystemClock.uptimeMillis() + delay)
```

rolandgriesser (changes requested):

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalyticsConfig.java
Tigraine: Yeah the name is not great - also it should go into the `CollectorConfig` that's specifically for configuring the behavior of the collector

collector/build.gradle
KulenovicInes: I needed it for testing private methods, now with refactoring it's not needed any more, so I removed it

collector/src/main/java/com/bitmovin/analytics/data/Backend.kt
KulenovicInes: Added `RetryCallback` interface that inherits this `okhttp3.Callback` interface

collector/src/main/java/com/bitmovin/analytics/utils/HttpClient.java
KulenovicInes: I removed two last setups, but this clients still is trying to resend data if I put only
`.retryOnConnectionFailure(false)` and I see duplicated samples in DB and if we put additional setting for connectTimeout then it's all working as expected.

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetrySample.kt
KulenovicInes: Moved this to `Comparator`

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
KulenovicInes: Created RetrySamples with generic

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
KulenovicInes: I overlooked it 🤦

collector/src/main/java/com/bitmovin/analytics/utils/HttpClient.java
```diff
@@ -19,15 +22,26 @@
     private static final String TAG = "HttpClient";
     private OkHttpClient client = null;
     private Context context;
+    private BitmovinAnalyticsConfig analyticsConfig;
 
-    public HttpClient(Context context) {
+    public HttpClient(Context context, BitmovinAnalyticsConfig analyticsConfig) {
```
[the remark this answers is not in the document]
KulenovicInes: Moved client outside and created factory for instantiating it

collector/src/main/java/com/bitmovin/analytics/utils/HttpClient.java
```diff
@@ -19,15 +22,26 @@
...
 
 
     public void post(String url, String postBody, final Callback callback) {
         if (client == null) {
-            client = new OkHttpClient();
+            if (analyticsConfig.isResendDataOnHttpTimeout()){
+                client = new OkHttpClient.Builder()
+
+                        .retryOnConnectionFailure(false)
+                        .connectTimeout(15, TimeUnit.SECONDS)
```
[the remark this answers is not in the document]
KulenovicInes: I think defaults are 10s. This client has its own retry process so it will try to resend failed requests. We need at least `connectTimeout(15, TimeUnit.SECONDS)` setting in combination with `retryOnConnectionFailure` to disable its own retry process. Otherwise I see duplicated samples in DB :(

collector/src/main/java/com/bitmovin/analytics/utils/Util.java
```diff
@@ -25,6 +25,8 @@
     public static final int ANALYTICS_QUALITY_CHANGE_COUNT_THRESHOLD = 50;
     public static final int ANALYTICS_QUALITY_CHANGE_COUNT_RESET_INTERVAL = 1000 * 60 * 60; // in milliseconds;
     public static final int REBUFFERING_TIMEOUT = 1000 * 60 * 2; // in milliseconds
+    public static final int MAX_RETRY_SAMPLES = 100;
+    public static final int MAX_RETRY_TIME =  300; //in seconds
```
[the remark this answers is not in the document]
KulenovicInes: For `Duration` I get `Call requires API level 26 (current min is 16)` :( So maybe we can add it later when we decide to upgrade API

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
KulenovicInes: Added generic parameter

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
KulenovicInes: I overlooked it 🤦

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
KulenovicInes: Well this entire method is in a lock, and I would really like to avoid additional complications with locking/unlocking with adding/removing data in queue.

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
KulenovicInes: That should be ok, because then this condition will not be executed. Then once first sample gets sent again `token` will be set to `null`. Then this `processQueuedSamples` method will be called again and `nextScheduleTime` will be from other sample that had same time, and that sample will be scheduled to be immediately sent.

KulenovicInes:
@Tigraine @rolandgriesser  I made a few changes from last PR
- added `RetryQueue` class which has responsibility of adding and removing data from queue
- replaced queue data structure from set to list because if scheduled times are identical (happens really rarely) set will just discard sample
- created `ClientFactory` for creating clients with config setup. AFAICS  all create clients calls are made from constructors so I don't think there will be some issues with multiple instances 😅 
- added some additional interfaces for backend and callbacks to reduce number of changed files

collector/src/main/java/com/bitmovin/analytics/utils/Util.java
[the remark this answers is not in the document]
Tigraine: @rolandgriesser should we rename the class or move it? Feels weird that these setting constants are living in a `Util.java` (should be a seperate PR)

collector/src/main/java/com/bitmovin/analytics/data/Backend.kt
```diff
@@ -0,0 +1,13 @@
+package com.bitmovin.analytics.data
+
+import com.bitmovin.analytics.retryBackend.RetryCallback
+
+interface Backend {
+    fun send(eventData: EventData)
+    fun sendAd(eventData: AdEventData)
+}
+
+interface CallbackBackend {
```
rolandgriesser, line 10: can you move this into a separate file?
Any reason you name this `RetryCallback`? This is just the normal callback as I can see.
What we can do though, is wrapping the `okhttp.Callback` into our own class (and then using some interface like the `RetryCallback`, just with a better naming as it's not related to `Retry`) - so we could decouple it from the `okhttp` (I think Daniel has mentioned that somewhere).
So you would have an interface like this e.g. (for now, we don't even need the success-path):
```
fun onFailure(e: Exception, cancel: () => Unit)
```
as those are the only 2 things we are interested in in the `RetryBackend`.

collector/src/main/java/com/bitmovin/analytics/data/BackendFactory.java
```diff
@@ -3,10 +3,18 @@
...
 public class BackendFactory {
 
     public Backend createBackend(BitmovinAnalyticsConfig config, Context context) {
-        return new HttpBackend(config.getConfig(), context);
+        HttpBackend httpBackend = new HttpBackend(config.getConfig(), context);
+        if(!config.getConfig().getTryResendDataOnFailedConnection() ){
+           return httpBackend;
+       }
+
+        return new RetryBackend(httpBackend, new Handler());
```
rolandgriesser, line 18: could this be a problem as we don't have access to the `Handler` instance anymore? Is unregistering stuff from the handler enough to not cause memory leaks?

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,105 @@
...
+    override fun sendAd(eventData: AdEventData) {
+        scheduleSample(RetrySample(eventData, 0, Date(), 0))
+    }
+
+    private fun scheduleSample(retrySample: RetrySample<Any>) {
+
+        if (retrySample.eventData is EventData) {
+            Log.d(TAG, "sending sample ${retrySample.eventData?.sequenceNumber} retry ${retrySample.retry}")
+            retrySample.eventData.retryCount = retrySample.retry
+            this.next.send(retrySample.eventData, object : RetryCallback {
```
rolandgriesser, line 40: I think you can assign the `RetryCallback` outside of the `if`, as it's not specific to `AdEventData` or `EventData`. So you only have to write it once. If we want to really overdo it, we could even create something like a `RetryableSample` interface that includes the `retryCount`, and can even share that code.
 The method would look something like this then:
`     private fun <T: RetryableSample>scheduleSample(retrySample: RetrySample<T>) {`

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,105 @@
...
+                }
+            })
+        }
+    }
+
+    @Synchronized
+    fun processQueuedSamples() {
+        try {
+            val nextScheduledTime = getNextScheduledTime()
+            if (nextScheduledTime != null) {
```
rolandgriesser, line 73: this should do the same and avoid nesting:
```suggestion
            if (nextScheduledTime == null || retryDateToken == nextScheduledTime) { 
            return
            }
```

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryQueue.kt
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Calendar
+import java.util.Date
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.Comparator
+import kotlin.math.pow
+
+class RetryQueue {
+    private val TAG = "RetryQueue"
+    private val lock = ReentrantLock()
+    private var retrySamplesSet = mutableListOf<RetrySample<Any>>()
```
rolandgriesser, line 15: why is this called set if it's a list?
wouldn't that work easier (so you wouldn't have to sort every time after adding a new sample)? 
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/sorted-set-of.html

rolandgriesser (changes requested):
Some minor things but looking really good already :)

Tigraine (commented):
Great work - looks much better already.
I did not review the `RetryBackend` class fully but @rolandgriesser already did .. But the structure looks much better already

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,105 @@
...
+                }
+
+                override fun onResponse(call: Call, response: Response) {
+                }
+            })
+        } else if (retrySample.eventData is AdEventData) {
+            retrySample.eventData.retryCount = retrySample.retry
+            this.next.sendAd(retrySample.eventData, object : RetryCallback {
+                override fun onFailure(call: Call, e: IOException) {
+                    if (e is SocketTimeoutException || e is ConnectException || e is StreamResetException || e is UnknownHostException) {
```
Tigraine, line 56: can the callback be shared between the two clauses? they are 1:1 the same

collector/src/main/java/com/bitmovin/analytics/data/Backend.kt
KulenovicInes: Moved to separate file, and changed interface in https://github.com/bitmovin/bitmovin-analytics-collector-android/pull/132/commits/f9d6c68a0c58cc6478feeacead3ae501992732d7

collector/src/main/java/com/bitmovin/analytics/data/BackendFactory.java
KulenovicInes: It was moved out for easier testing... And new Handler instance is created in factory inside `retryBackend` constructor But that class has destructor so I assume it should be fine

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryQueue.kt
KulenovicInes: Tried to implement set with comparator and it's really hard and messy to be able to add `retrySamples` with diffrent scheduled time and still have a equality check for event object which is generic.... All the time I got either samples being discarded because of same scheduled time or if I managed to add them into the set then `remove` method stops working because it can't determine which is correct sample to remove (and this causes literary hundreds of same samples being sent to backend in a few seconds in a infinite loop). Things like PriorityQueue would require upgrade of API verson :(
So simplest and most functional and safe solution is to use lists

collector/src/main/java/com/bitmovin/analytics/data/BackendFactory.java
```diff
@@ -3,10 +3,18 @@
...
+
 
 public class BackendFactory {
 
     public Backend createBackend(BitmovinAnalyticsConfig config, Context context) {
-        return new HttpBackend(config.getConfig(), context);
+        HttpBackend httpBackend = new HttpBackend(config.getConfig(), context);
+        if(!config.getConfig().getTryResendDataOnFailedConnection() ){
+           return httpBackend;
+       }
```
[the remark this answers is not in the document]
KulenovicInes: 🙈  https://github.com/bitmovin/bitmovin-analytics-collector-android/pull/132/commits/bd35a27a177c590c4782d8e91366465e132c7fd5

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
KulenovicInes: Yes it can.. move it out in https://github.com/bitmovin/bitmovin-analytics-collector-android/pull/132/commits/f9d6c68a0c58cc6478feeacead3ae501992732d7

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryCallback.kt
```diff
@@ -0,0 +1,14 @@
+package com.bitmovin.analytics.retryBackend
+
+import java.io.IOException
+import okhttp3.Call
+import okhttp3.Callback
+import okhttp3.Response
+
+interface RetryCallback : Callback {
+    override fun onFailure(call: Call, e: IOException) {
```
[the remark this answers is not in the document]
KulenovicInes: Interfaces are changed in https://github.com/bitmovin/bitmovin-analytics-collector-android/pull/132/commits/f9d6c68a0c58cc6478feeacead3ae501992732d7
so now it's fully separated from okhttp3

collector/src/main/java/com/bitmovin/analytics/data/BackendFactory.java
Tigraine: Since nobody has a reference to it the GC should collect it. So if only RetryBackend + Handler reference each other they end up outside the alive set so I don't think there can be a memory issue

collector/src/main/java/com/bitmovin/analytics/data/HttpBackend.java
```diff
@@ -18,28 +26,61 @@ public HttpBackend(CollectorConfig config, Context context) {
...
                 eventData.getStartupTime(),
                 eventData.getVideoStartupTime(),
-                eventData.getBuffered(), 
+                eventData.getBuffered(),
                 eventData.getAudioLanguage()));
-        this.httpClient.post(analyticsBackendUrl, DataSerializer.serialize(eventData), null);
+        this.httpClient.post(analyticsBackendUrl, DataSerializer.serialize(eventData), new Callback() {
+            @Override
+            public void onFailure(Call call, IOException e) {
+                if(callback != null) {
```
Tigraine, line 55: Nitpick: you could invert the null for early return

collector/src/main/java/com/bitmovin/analytics/data/HttpBackend.java
```diff
@@ -18,28 +26,61 @@ public HttpBackend(CollectorConfig config, Context context) {
...
+    public void sendAd(AdEventData eventData, OnFailureCallback callback) {
         Log.d(TAG, String.format("Sending ad sample: %s (videoImpressionId: %s, adImpressionId: %s)",
                 eventData.getAdImpressionId(),
                 eventData.getVideoImpressionId(),
                 eventData.getAdImpressionId()));
-        this.httpClient.post(adsAnalyticsBackendUrl, DataSerializer.serialize(eventData), null);
+        this.httpClient.post(adsAnalyticsBackendUrl, DataSerializer.serialize(eventData), new Callback() {
+            @Override
+            public void onFailure(Call call, IOException e) {
+                if(callback !=null) {
```
Tigraine, line 75: invert and formatting `!=null`

Can you run the formatter (either inside intelliJ or ktlint / coveo?)

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,87 @@
...
+import com.bitmovin.analytics.data.CallbackBackend
+import com.bitmovin.analytics.data.EventData
+import java.net.ConnectException
+import java.net.SocketTimeoutException
+import java.net.UnknownHostException
+import java.util.Date
+import kotlin.Exception
+import okhttp3.internal.http2.StreamResetException
+
+class RetryBackend(private val next: CallbackBackend, val handler: Handler) : Backend {
```
Tigraine, line 17: maybe there is a better name for the `handler` field. I am not sure what this is actually doing but it looks like it's a common thing in Android.,

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,87 @@
...
+                }
+            }
+        }
+
+        if (retrySample.eventData is EventData) {
+            Log.d(TAG, "sending sample ${retrySample.eventData?.sequenceNumber} retry ${retrySample.retry}")
+            retrySample.eventData.retryCount = retrySample.retry
+            this.next.send(retrySample.eventData, callback)
+        } else if (retrySample.eventData is AdEventData) {
+            retrySample.eventData.retryCount = retrySample.retry
```
Tigraine, line 48: the else block is not logging

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,87 @@
...
+            if (nextScheduledTime != null || retryDateToken != nextScheduledTime) {
+                if (retryDateToken != null) {
+                    handler.removeCallbacks(processSampleRunnable, retryDateToken)
+                }
+                retryDateToken = nextScheduledTime
+                val delay = maxOf((nextScheduledTime?.time ?: 0) - Date().time, 0) // to prevent negative delay
+                handler.postAtTime(processSampleRunnable, retryDateToken, SystemClock.uptimeMillis() + delay)
+            }
+        } catch (e: Exception) {
+            Log.d(TAG, "processQueuedSamples ${e.message}")
```
Tigraine, line 66: better error message would be useful here (at least that an exception happened).
Also i think you can pass the exception so you get a stacktrace etc. This should not be `Log.d` but `Log.e`
https://developer.android.com/reference/android/util/Log#e(java.lang.String,%20java.lang.String,%20java.lang.Throwable)

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,87 @@
...
+    private val processSampleRunnable = Runnable {
+        try {
+            val retrySample = retryQueue.getNextSampleOrNull()
+            if (retrySample != null) {
+                scheduleSample(retrySample)
+            }
+            retryDateToken = null
+            processQueuedSamples()
+        } catch (e: Exception) {
+            Log.d(TAG, "processSampleRunnable ${e.message}")
```
Tigraine, line 79: `Log.e` https://developer.android.com/reference/android/util/Log#e(java.lang.String,%20java.lang.String,%20java.lang.Throwable)

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,87 @@
...
+                retryDateToken = nextScheduledTime
+                val delay = maxOf((nextScheduledTime?.time ?: 0) - Date().time, 0) // to prevent negative delay
+                handler.postAtTime(processSampleRunnable, retryDateToken, SystemClock.uptimeMillis() + delay)
+            }
+        } catch (e: Exception) {
+            Log.d(TAG, "processQueuedSamples ${e.message}")
+        }
+    }
+
+    private val processSampleRunnable = Runnable {
```
Tigraine, line 70: What is this method doing? A better name or some JavaDoc would be great

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,87 @@
...
+        try {
+            val retrySample = retryQueue.getNextSampleOrNull()
+            if (retrySample != null) {
+                scheduleSample(retrySample)
+            }
+            retryDateToken = null
```
Tigraine, line 76: What is the reasoning for naming the field `token`?

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,87 @@
...
+            processQueuedSamples()
+        } catch (e: Exception) {
+            Log.d(TAG, "processSampleRunnable ${e.message}")
+        }
+    }
+
+    protected fun finalize() { // destroys an instance of RetryBackend
+        Log.d(TAG, "destroy")
```
Tigraine, line 84: Shouldn't this be `finalize?

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryQueue.kt
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    fun addSample(retrySample: RetrySample<Any>) {
+
+        try {
+            lock.lock()
+            retrySample.retry++
+
+            val backOffTime = minOf(2.toDouble().pow(retrySample.retry).toInt(), 64)
```
Tigraine, line 33: `64` should be a constant field with a name

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryQueue.kt
```diff
@@ -0,0 +1,81 @@
...
+            retrySample.totalTime += backOffTime
+            // more than 5min in queue
+            if (retrySample.totalTime < Util.MAX_RETRY_TIME) {
```
Tigraine, line 36: can we move the constant to this class and not to the `Util` class? Feels weird that we have a configuration constant in a `Util` class

Tigraine, line 36: with early return we could reduce nesting here

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryQueue.kt
```diff
@@ -0,0 +1,81 @@
...
+                retrySample.scheduledTime = Calendar.getInstance().run {
+                    add(Calendar.SECOND, backOffTime)
+                    time
+                }
+
+                if (retrySamplesList.size > getMaxSampleNumber()) {
```
Tigraine, line 42: wouldn't this allow the `maxSampleNumber + 1`?

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryQueue.kt
```diff
@@ -0,0 +1,81 @@
...
+            Log.d(TAG, "addSample ${e.message}")
+        } finally {
+            lock.unlock()
+        }
+    }
+
+    fun getNextSampleOrNull(): RetrySample<Any>? {
+        try {
+            lock.lock()
+            val retrySample = retrySamplesList.firstOrNull { it.scheduledTime <= Date() }
```
Tigraine, line 60: abstract out the `Date()`

Tigraine (changes requested):

collector/src/main/java/com/bitmovin/analytics/data/HttpBackend.java
KulenovicInes: fixed in https://github.com/bitmovin/bitmovin-analytics-collector-android/pull/132/commits/8ed6142f98230909ac19f34acadc4ad135cae400

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
KulenovicInes: renamed handler and runnable in https://github.com/bitmovin/bitmovin-analytics-collector-android/pull/132/commits/0b5f3e8dcbe9116a274fd705ccfda193d020ff56
are `scheduleSampleHandler` and `sendNextSampleRunnable` better?

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
KulenovicInes: changed logs in https://github.com/bitmovin/bitmovin-analytics-collector-android/pull/132/commits/479546852b7e47cf387dec0b6ff9b8b7f909c831

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryQueue.kt
KulenovicInes: moved constants from util https://github.com/bitmovin/bitmovin-analytics-collector-android/pull/132/commits/626828002ef8e7411471c2172d47085a68222dc0
also added new constant `MAX_BACKOFF_INTERVAL = 64` and fixed its value (bacause I'm blind 🙈 )
https://github.com/bitmovin/bitmovin-analytics-collector-android/pull/132/commits/f93573e337d8ec4813d97988c1890eedca2b8b5e

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
KulenovicInes: Logic for re-sending samples uses handlers for scheduling execution of logic for getting and sending next sample.
https://developer.android.com/reference/android/os/Handler#postAtTime(java.lang.Runnable,%20java.lang.Object,%20long)

Method used `postAtTime` receives 3 parameters:
- runnable (interface that wraps code which should be executed)
- token (an instance which can be used to cancel runnable, can be any object)
- uptimeMillis (when runnable should be executed)

So I wanted to indicate in variable name as what this variable is used. I can also rename it to something like `nextScheduledExecutionTime`?

Tigraine: But this is not a time then but rather a `cancelToken`? Or are you just using the same object for storing the executionTime + the cancel Token ?

KulenovicInes: Well both... I save next scheduled time in this variable so I can check if there is some sample which should be sent before, if there is then I use it as cancel token.

KulenovicInes:
> The error message is still too short IMO..
> It would be better if it at least said something descriptive: "processedQueueSamples threw an unexpected Exception" .. because unless you inspect the exception stacktrace there is really no explanation why a ERROR logline is here.

Changed in https://github.com/bitmovin/bitmovin-analytics-collector-android/pull/132/commits/2a37ccdbaa1edccba5b3284e815ba3b9587a9cd6

collector/src/main/java/com/bitmovin/analytics/BitmovinAnalytics.java
```diff
@@ -405,4 +405,4 @@ public void message(@NotNull String message) {
             }
         }
     };
-}
+  }
```
sabe25, line 408: Do we have formater in this project?

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryBackend.kt
```diff
@@ -0,0 +1,89 @@
...
+            processQueuedSamples()
+        } catch (e: Exception) {
+            Log.e(TAG, "processSampleRunnable() threw an unexpected exception: ${e.message}", e)
+        }
+    }
+
+    protected fun finalize() { // destroys an instance of RetryBackend
+        Log.d(TAG, "finalize")
+        scheduleSampleHandler.removeCallbacksAndMessages(null) // removes all callbacks and messages
+    }
```
sabe25, line 88: https://stackoverflow.com/questions/48623441/destructor-in-kotlin-programming-language
I found this regarding the finalize method in java/kotlin...
Any thoughts/experiences with this?

Tigraine: Problem @sabe25 is that this is not being used inside one small `using` statement but rather for a longer kind and when it gets collected we need to make sure these callbacks get unregistered too. I think this usage is safe here

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryQueue.kt
```diff
@@ -0,0 +1,84 @@
...
+import kotlin.math.pow
+
+class RetryQueue {
+    private val TAG = "RetryQueue"
+    private val lock = ReentrantLock()
+    private val MAX_RETRY_TIME = 300 // in seconds
+    private val MAX_RETRY_SAMPLES = 100
+    private val MAX_BACKOFF_INTERVAL = 64
+
+    fun getMaxSampleNumber() = MAX_RETRY_SAMPLES
```
Tigraine, line 17: why does this need to be a fun and not just a private val?

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryQueue.kt
```diff
@@ -0,0 +1,84 @@
...
+            }
+
+            retrySample.scheduledTime = Calendar.getInstance().run {
+                add(Calendar.SECOND, backOffTime)
+                time
+            }
+            if (retrySamplesList.size >= getMaxSampleNumber()) {
+                val removeSample = retrySamplesList.last()
+                retrySamplesList.remove(removeSample)
+                Log.d(TAG, "removed sample")
```
Tigraine, line 49: more extensive logging would be helpful (`removed oldest sample due to queue being over capacity of %d` or something)

Tigraine (approved):
LGTM but two smaller style comments you might want to consider.
Good job!

collector/src/main/java/com/bitmovin/analytics/retryBackend/RetryQueue.kt
KulenovicInes: I did it like this so I can change it to some smaller value in tests.
Other option I can think of is to use `public var MAX_BACKOFF_INTERVAL` and then change it 🤔

Tigraine: ok no it's fine this way - just looked odd

PR #135, opened 2021-01-25
AN1988 - Feature/concatenating media source test

rolandgriesser:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-1988
- Description: If the player is already loading when attach the collector (e.g. attaching the collector when the index within a `ConcatenatingMediaSource` changes, to avoid tracking a bumper ad), we now directly transition into the `STARTUP` state, as we got stuck in `READY` till the first `PLAYED`/`PAUSED` event otherwise (for the Bitmovin player, this behaviour was already implemented in the `BitmovinSdkAdapter.checkAutoplayStartup()`)

### Checklist

- [x] Updated CHANGELOG.md


sabe25 (approved):

collector-exoplayer/src/main/java/com/bitmovin/analytics/exoplayer/ExoPlayerAdapter.java
```diff
@@ -123,6 +123,25 @@ public void init() {
...
+    private void checkAutoplayStartup() {
+        int playbackState = exoplayer.getPlaybackState();
+        boolean playWhenReady = exoplayer.getPlayWhenReady();
+        if(playbackState == Player.STATE_BUFFERING)
+        {
+            if (playWhenReady != this.isPlaying && playWhenReady) {
+                Log.d(TAG, "Collector was attached while media source was already loading, transitioning to startup state.");
+                // with autoplay enabled the player first enter here and start buffering for the video with playWhenReady = true
+                this.isPlaying = true;
+                startup(getPosition());
```
Tigraine, line 142: how will this impact the startup measurements? do we want to collect startup-times for these cases?

Tigraine (approved):

PR #140, opened 2021-02-18
Feature/segment tracking testing

rolandgriesser:
### Work
- Fixes: https://bitmovin.atlassian.net/browse/AN-2075
- Description: Added refactoring and tests around the new FeatureManager and the corresponding features

### Checklist

- [ ] Updated CHANGELOG.md: NOT NECESSARY

# Hint

- Reviewing might be easier if you compare this branch to the last tag, as you can see the whole features working together then:
https://github.com/bitmovin/bitmovin-analytics-collector-android/compare/v1.22.0...feature/segment-tracking-testing


collector/src/test/java/com/bitmovin/analytics/features/errordetails/ErrorDetailBackendTests.kt
```diff
@@ -0,0 +1,70 @@
...
+        assertThat(d1.segments?.size).isLessThanOrEqualTo(2)
+        assertThat(d2.segments?.size).isLessThanOrEqualTo(2)
+        assertThat(d3.segments?.size).isNull()
+    }
+
+    @Test
+    fun testWontPostWhileDisabled() {
+        val backend = ErrorDetailBackend(mockk())
+        backend.send(getErrorDetail(0))
+        verify(exactly = 0) { anyConstructed<HttpClient>().post(any(), any(), any()) }
```
sabe25, line 38: `anyConstructed` how awesome is that? 😅

sabe25 (approved):
LGTM 
Nice amount of Tests 👍

collector-bitmovin-player/build.gradle
```diff
@@ -57,6 +57,8 @@ dependencies {
 		api "com.bitmovin.analytics:collector:${rootProject.ext.version}"
 	}
 	implementation "com.bitmovin.player:playercore:${rootProject.ext.playercoreVersion}"
+
+	testImplementation "io.mockk:mockk:1.10.6"
```
Tigraine, line 61: do we really need yet another mock library? what is this for? don't we already have mockito?

Tigraine (commented):
Can we not use mockito like we do everywhere else? I am not sure if it's good to use another mocking library which nobody has any experience with.
Might lead to a lot of confusion

collector-bitmovin-player/build.gradle
rolandgriesser: There were a lot of things that didn’t work too well with mockito and Kotlin. And with this library there is good Generic support and mocking/verifying works pretty well and cleanly. Mostly, being able to use the mockKConstructor makes a big difference - our collector wasn’t build with having tests in mind, so it’s really hard to pass any mocks in a lot of places.

Svarozic: I support decision to use that library.

In other projects we are using `mockito-kotlin` but that is just wrapper arround `mockito` to solve compilation issues (you can still have runtime mocking issues). I checked some discussions on web and even author of `mockito-kotlin` would recoment to use that library:
https://www.reddit.com/r/androiddev/comments/9us0a0/mockk_vs_mockitokotlin/

@rolandgriesser I would set this dep to all modules that have unit tests, and share version number in root `build.gradle` file , so we have modules in sync (as in past when I first time touched project, I had an issue with different lib versions in different module in this project)

